Issue #330, opened 2014-09-18
Adding not-autodetected account result in 400


LEDfan:
When trying to add an account with advanced configuration the request fails:
![naamloos](https://cloud.githubusercontent.com/assets/2996275/4319264/25959e16-3f2a-11e4-8376-899cdd347b4c.png)

Tested on PHP 5.4 & 5.3. No errors in the log.

PR #333, opened 2014-09-18
if manual setup is open, close it after successful connection, fix #331

### LINKED ISSUE #331: Failed advanced account creation breaks interface
When adding an account using the advanced configuration options, the top password box disappears since there are two separate IMAP and SMTP passwords instead. However, if creating that account fails, the advanced settings dialog closes, but the top password box does not return. Opening advanced configuration brings that password box back, at least until the advanced configuration is closed again.

A fix would be to not close advanced settings when adding the account fails.

jancborchardt:
This moves the manual setup toggle code into a function. Also fixes #331 in that the manual setup area is only closed on successful connection.

Please test @Xenopathic @zinks- @PoPoutdoor 

And have a look at the code @DeepDiver1975 ;)

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -517,6 +517,19 @@ var Mail = {
 		setFolderInactive:function (accountId, folderId) {
 			$('.mail_folders[data-account_id="' + accountId + '"] li[data-folder_id="' + folderId + '"]')
 				.removeClass('active');
+		},
+
+		toggleManualSetup:function() {
+			$('#mail-setup-manual').slideToggle();
+			$('#mail-imap-host').focus();
+			if($('#mail-address').parent().prop('class') === 'groupmiddle') {
+				$('#mail-password').slideToggle(function() {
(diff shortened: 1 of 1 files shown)


Xenopathic:
The manual setup no longer closes if creating the account failed, which is good, but I seem to be unable to test a successful manual setup due to other issues (I'm getting 400 Bad Request). Is it necessary to explicitly close the manual setup if creating the account succeeded though? I thought that the display immediately changes to something else anyway, so surely it is unneeded and will never be seen?

Xenopathic:
Marking as priority as it is very user-unfriendly and confusing. Please rebase :smile:

jancborchardt:
Manually rebased the current state in https://github.com/owncloud/mail/pull/504

jancborchardt:
@Xenopathic btw, the 400 bad request might be related to https://github.com/owncloud/mail/issues/330 – a fix is available at https://github.com/owncloud/mail/pull/436, can you test that?

PR #428, opened 2015-03-08
show 'New Message' button on emtpy accounts

### LINKED ISSUE #427: "New Message" button disabled when using an empty account
I discovered that the mail app does not allow sending any mails as long as the current account does not have mails yet:
oc_mail_sent1

To be able to send new messages, I have to send an initial mail to this account. Then, the button is enabled again:

oc_mail_sent2

wurstchristoph:
As described [here](https://github.com/owncloud/mail/issues/427), 'New Messages' has some strange behavior on emtpy accounts:
- it is visible and active when the app loads
- it gets disabled when you click it
- when a user clicks on a folder again, the button does not become active again

fixes #427

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -236,6 +236,11 @@ var Mail = {
 							Mail.State.currentFolderId = folderId;
 							Mail.UI.setMessageActive(null);
 							$('#mail_messages').removeClass('icon-loading');
+							
+							// Fade out the message composer
+							$('#mail_new_message').prop('disabled', false);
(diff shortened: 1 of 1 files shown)


jancborchardt:
Indeed weird that it was only in the openMessage function. Thank you very much for the fix! :)

If you want to help even more, check out some of the [issues for our 0.1 milestone](https://github.com/owncloud/mail/issues?q=is%3Aopen+is%3Aissue+milestone%3A0.1).

PR #429, opened 2015-03-08
hide 'Load More' button if message count smaller than pagination limit

### LINKED ISSUE #388: 'Load more' button shown even if all messages are displayed
The 'Load more' button should not be displayed if the mailbox's total message count is less than our pagination limit.

wurstchristoph:
fixes #388


jancborchardt:
Nice, works as well! :) :+1:

Issue #425, opened 2015-02-18
[oC8] Doesn’t load emails


jancborchardt:
With ownCloud 8, the web interface does not load emails (»Error while loading messages«) and I get this error in the owncloud.log:
```
{"reqId":"ioI64F0dCNUlFmx7hJWq","remoteAddr":"127.0.0.1","app":"index","message":"Exception: {"Message":"IMAP error reported by server.","Code":0,"Trace":"#0 /home/jan/owncloud-dev/apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/Socket.php(4206): Horde_Imap_Client_Socket->_getLine(Object(Horde_Imap_Client_Interaction_Pipeline))\
#1 /home/jan/owncloud-dev/apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/Socket.php(4136): Horde_Imap_Client_Socket->_sendCmdChunk(Object(Horde_Imap_Client_Interaction_Pipeline), Array)\
#2 /home/jan/owncloud-dev/apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/Socket.php(1659): Horde_Imap_Client_Socket->_sendCmd(Object(Horde_Imap_Client_Interaction_Pipeline))\
#3 /home/jan/owncloud-dev/apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/Base.php(1687): Horde_Imap_Client_Socket->_status(Array, 1)\
#4 /home/jan/owncloud-dev/apps/mail/lib/mailbox.php(125): Horde_Imap_Client_Base->status(Object(Horde_Imap_Client_Mailbox), 1)\
#5 /home/jan/owncloud-dev/apps/mail/lib/mailbox.php(129): OCA\Mail\Mailbox->getStatus(1)\
#6 /home/jan/owncloud-dev/apps/mail/lib/mailbox.php(67): OCA\Mail\Mailbox->getTotalMessages()\
#7 /home/jan/owncloud-dev/apps/mail/lib/controller/messagescontroller.php(65): OCA\Mail\Mailbox->getMessages(0, 20)\
#8 [internal function]: OCA\Mail\Controller\MessagesController->index(0, 20)\
#9 /home/jan/owncloud-dev/lib/private/appframework/http/dispatcher.php(158): call_user_func_array(Array, Array)\
[...]

PR #431, opened 2015-03-18
Don't double encode

LukasReschke:
Meanwhile we do it already in core...

Partially fixes https://github.com/owncloud/mail/issues/425

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -228,7 +228,7 @@ var Mail = {
 			} else {
 				$.ajax(
 					OC.generateUrl('apps/mail/accounts/{accountId}/folders/{folderId}/messages',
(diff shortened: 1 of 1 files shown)


jancborchardt:
@LukasReschke should there be visible changes observable? Cause I don’t see any change yet. :) I assume it’s only to fix some log errors for now?

LukasReschke:
Try to use Gmail with it and you will see why this is needed.

That said, as stated in the referenced issue, if the app does not work at all for you use an older PHP version or wait until PHP fixed their regression :see_no_evil:

jancborchardt:
Ohlala :O there’s no way we can fix it? Or is it just a matter of free beers for Lukas? :D

PR #434, opened 2015-03-27
autoconfig logging added

wurstchristoph:
looking at issue #330, I found out that the error message ```'Creating account failed: '``` was passed to translation with the exception message, resulting in an empty message string in the json response.

I also added some logging for further issues with autconfig.


jancborchardt:
@DeepDiver1975 @Gomez @zinks- for you to review ;)

DeepDiver1975:
:+1:

PR #435, opened 2015-03-27
Notes on scratch install

Gomez:
Easy way for new contributer


jancborchardt:
You sure we should list that here? Maybe just a link to the docs instead? cc @DeepDiver1975 @wurstchristoph @zinks-  what do you think?

jancborchardt:
Talked to @DeepDiver1975 as well – we don’t really care how people set up their ownCloud. A link to the installation documentation would be enough in this case. Do you want to add that?

PR #439, opened 2015-03-29
decrease unread counter

### LINKED ISSUE #246: Reading a message does not decrease "unread" counter
The count of a folder's unread messages (displayed in the side bar) doesn't change untill the account is reloaded.

wurstchristoph:
changeMessageFlags did not have the old attribute state anymore, so the evaluation logic was moved to the changeFlags listener function

With this change, the unread counter seems to work again. However, I'm sure there's possibly a better solution since I am no JavaScript expert ;-)

fixes #246

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -105,8 +105,8 @@ var Mail = {
 			});
 			Mail.State.folderView.render();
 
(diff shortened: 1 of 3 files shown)


jancborchardt:
Works great! :) One more step towards 0.1

PR #441, opened 2015-03-30
Correctly decode transfer encoding on message body

### LINKED ISSUE #399: Blank mail content: iconv illegal character
Another spam trigger OC error:
~~~
Error 	PHP 	iconv(): Detected an illegal character in input string at /${OC_ROOT}/apps/mail/lib/message.php#443 
~~~
Mail header
~~~
Subject:  Re:
MIME-Version: 1.0
Content-Type: text/plain; charset="us-ascii"
......
Content-Transfer-Encoding: quoted-printable
~~~
In mail header indicate **charset="us-ascii"**, but the correct encoding is **windows-1252**. The illegal character is converted to UTF-8 as XML &#146, which is display as **&#8217;** (XML &#8217) if decoded as windows-1252 and converted to UTF8.

Xenopathic:
Don't trust the remote server to decode the MIME transfer type, it doesn't work for Dovecot at least. Instead, decode the body part ourselves, like attachments get decoded.

With this, I can now read base64 transfer encoded emails!

Fixes #399

### DIFF
--- a/lib/message.php
+++ b/lib/message.php
@@ -23,6 +23,7 @@
 namespace OCA\Mail;
 
 use Horde_Imap_Client;
+use Horde_Imap_Client_Data_Fetch;
 use OCA\Mail\Service\Html;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\Util;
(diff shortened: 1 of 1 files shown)


jancborchardt:
@LukasReschke @PoPoutdoor can you review this? :)

clementhk:
https://github.com/horde/horde/blob/220b3625577293b08c20731fd1d914395b8f1ef0/framework/Mime/lib/Horde/Mime/Part.php#L1837

It's either fixed in master or something else went wrong, would be cool to find out the source as this is more like a hack, and maybe should have a comment that says it is a hack.

jancborchardt:
@clementhk if you want to look into this that would be cool! :) Btw, you’re also very welcome in our IRC channel #owncloud-mail (on freenode IRC).

jancborchardt:
@Xenopathic @clementhk @PoPoutdoor so what should we do about this specific fix here? Pull it in or not? :)

Xenopathic:
Well it is definitely broken in master, and definitely fixed on this PR. Judge the code for yourself, but IMHO it makes sense :smile:

PR #447, opened 2015-04-08
Fix Content Security Policy for frame-src

Spacefish:
To make HTML mails work again in Owncloud 8.1


DeepDiver1975:
@LukasReschke THX

LukasReschke:
:+1: – though the fact that the tests do still pass mean we probably have no unit tests for this function :)

DeepDiver1975:
:+1:

Spacefish:
@LukasReschke  yeah looks like it :) i saw Core has a lot of Unit Tests arround Content-Security things. But Mail doesn´t

Issue #448, opened 2015-04-09
»Message sent« notification is buggy


jancborchardt:
When sending an email, the »message sent« notification below the input field is a bit buggy. When clicking the send button it shows `[object Object]` while loading and then when it’s sent it will say:

`Message sent to &amp;quot;Jan-Christoph Borchardt&amp;quot; &amp;lt;jan@bla.bla&amp;gt;`

Seems something is escaped wrongly. cc @DeepDiver1975 @LukasReschke @wurstchristoph

LukasReschke:
We need to re-evaluate all usages of `t()` and remove the manual escaping, it is now done automatically.

For example in `message: t('mail', 'Message sent to {Receiver}', {Receiver: escapeHTML(to.val())})`  the `escapeHTML` has to get removed.

jancborchardt:
Fixed by https://github.com/owncloud/mail/pull/458, thanks to @wurstchristoph @DeepDiver1975 :)

PR #451, opened 2015-04-09
remove double escaping

ChristophWurst:
as described in #448, the message shown when mails are sent was double escaped.
To prevent the `[object Object]` output, I changed `{}` to an empty string - does this line make any sense at all? should the app print something about sending the mail?

This is only a fix for sending new mails. If you reply to a mail the `to` value is extracted from the input field, which is already escaped, so the double escaping occurs again.

Does anybody know a good solution to fix replying too? @jancborchardt @LukasReschke

### DIFF
--- a/js/views/sendmail.js
+++ b/js/views/sendmail.js
@@ -1,4 +1,4 @@
-/* global Backbone, Handlebars, models, escapeHTML */
+/* global Backbone, Handlebars, models */
 
 var views = views || {};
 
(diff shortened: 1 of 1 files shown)


jancborchardt:
@LukasReschke can you comment here and/or help fixing it? ;)

jancborchardt:
Cool! This fixes the `[object Object]` stuff, but the escaping issue is still there. (For both new messages and replies.) @DeepDiver1975 is looking into it. :)

jancborchardt:
Follow-up at https://github.com/owncloud/mail/pull/458:)

PR #452, opened 2015-04-09
highlight only one folder at a time

### LINKED ISSUE #440: Several folders highlighted when clicking through the list fast
When you click on different folders faster than they can load, the sidebar can look like this:
mail-folders
(In this case the »Sent« folder is open since it was clicked last.)

To fix this, we should remove the highlight on any other folder when clicking on a new folder. cc @wurstchristoph @zinks-

wurstchristoph:
```setFolderInactive``` removed as it is not necessary anymore

fixes #440

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -200,7 +200,6 @@ var Mail = {
 
 		loadMessages:function (accountId, folderId, noSelect) {
 			// Set folder active
-			Mail.UI.setFolderInactive(Mail.State.currentAccountId, Mail.State.currentFolderId);
 			Mail.UI.setFolderActive(accountId, folderId);
 			Mail.UI.clearMessages();
 			$('#mail_messages')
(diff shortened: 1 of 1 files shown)


jancborchardt:
Will check it out right now! Does this also fix the follow-up issue you mentioned as comment in #440 about the message list loading for every folder you click? Or would that be a separate PR?

jancborchardt:
(Answering myself: No it doesn’t fix that second issue. :) But the highlighting fix works!

@wurstchristoph do you want to fix the other issue in this PR as well? Because it kind of belongs together it seems. Otherwise the UI is more polished than the requests. ;)

jancborchardt:
Oh, it seems that this pull request (in comparison to master) results in sending 3 or 4 emails instead of one. ;)

(You probably will see it in your inbox as I sent you an email just now.)

wurstchristoph:
Are you sure that multiple sent emails are caused by this PR? It doesn't change anything but the UI.

Regarding the second issue: To be honest, I have no idea on how to deal with those concurrent ajax calls, therefore I cannot fix it (for now)... :disappointed:

jancborchardt:
Ok, I’ll have to test it again. Weirdly enough it happened, and didn’t happen when I switched back to master. But maybe it was something else.

jancborchardt:
The multiple mail sending must have been something else. :+1: here. :)

PR #456, opened 2015-04-10
NoCSRFRequired on ProxyController::proxy()

DeepDiver1975:
@LukasReschke


jancborchardt:
This fixes a bunch of images not being loaded in HTML mails. :+1: 

@LukasReschke please review.

LukasReschke:
Yes. That works. I think we can harden it further using some CSP policies. I need to take a closer look at the new in-draft RFC again ;-) :+1:

PR #464, opened 2015-04-11
hide technical attachments like signature.asc and smime.p7s, fix #126

### LINKED ISSUE #126: If attachment is signature.asc or smime.p7s, don’t show it
Attachments like `signature.asc`, `smime.p7s` (and other technical things, if there’s more) should not be shown in the attachment list.

jancborchardt:
Attachments like signature.asc and smime.p7s will just not be shown in the attachment list. Fix #126
Also if that attachment was the only one, the attachment indicator will not show either.

Please review @DeepDiver1975 @wurstchristoph – there’s probably a more elegant way of defining which files to skip instead of hardcoding them twice. ;)
Also, do you know of any other technical files which show up as attachments which should be hdiden? cc @Gomez @zinks-

### DIFF
--- a/lib/message.php
+++ b/lib/message.php
@@ -208,13 +208,19 @@ class Message {
 			 * @var \Horde_Mime_Part $p
 			 */
 			$filename = $p->getName();
-			if (!is_null($p->getContentId())) {
+
+			if(!is_null($p->getContentId())) {
 				continue;
 			}
-			if (isset($filename)) {
-				return true;
+			if(isset($filename)) {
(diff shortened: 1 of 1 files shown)


jancborchardt:
@LukasReschke or are these attachments used for something?

DeepDiver1975:
> there’s probably a more elegant way of defining which files to skip instead of hardcoding them twice. ;)

yes :speak_no_evil: -  a121229

jancborchardt:
Great stuff, thank you! :+1:

PR #470, opened 2015-04-13
directly change star state in the interface for quick feedback

jancborchardt:
Currently we show a little loading spinner while the message is being starred. This results in perceived slowness of the app.

So instead, we just lie and change the status directly. In the off-case there is an error it will just change back.

Please review @DeepDiver1975 @wurstchristoph @Gomez @zinks- @LukasReschke


wurstchristoph:
Looks good to me :+1:

PR #471, opened 2015-04-13
Fixing download of attachments where the filename is within the content-type and not the content-disposition 

### LINKED ISSUE #438: Attachment name wrong on download & save to files (only on one mail server)
When downloading attachments or saving them to files, the name is not preserved. They are just named »(2).« when saving to files and just »2« when downloading.

cc @DeepDiver1975 @wurstchristoph any idea? Can you reproduce?

DeepDiver1975:
fixes #438

@jancborchardt


jancborchardt:
Great stuff. Tested and it works nicely. :+1: thanks a lot!

PR #472, opened 2015-04-13
Implement search

DeepDiver1975:
@jancborchardt

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -1,6 +1,6 @@
 /* global Handlebars, Marionette, relative_modified_date, formatDate, humanFileSize, views */
 var Mail = {
-	State:{
+	State: {
 		currentFolderId: null,
 		currentAccountId: null,
 		currentMessageId: null,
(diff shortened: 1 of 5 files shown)


jancborchardt:
Nice, great start! Just one issue: No other folder than the inbox of the first account can be opened.

The other folders show »No messages in this folder!« and this error in the log:
```
{"reqId":"g\/pFb4D5TJfYbCPtl6Gf","remoteAddr":"127.0.0.1",
"app":"PHP","message":"Undefined index: text at \/home\/jan\/owncloud\/apps\/mail\/vendor\/pear-pear.horde.org\/Horde_Imap_Client\/Horde\/Imap\/Client\/Search\/Query.php#243",
"level":0,"time":"2015-04-13T22:11:10+00:00","method":"GET",
"url":"\/owncloud\/index.php\/apps\/mail\/accounts\/2\/folders\/SU5CT1g%3D\/messages"}
```

jancborchardt:
Nice!!!

Just one thing left now: If you have multiple accounts, the folder highlight is a bit messed up. Each account always has a folder highlighted as active. So even if you only have the sent folder of account 2 open, the last folder you opened in account 1 will also be highlighted in the sidebar.

jancborchardt:
And second thing: The »load more« button is missing.

jancborchardt:
Another thing: What seems to be searched is the sender and the message body. What is _not_ searched though is the subject and people in cc. Can we add that?

jancborchardt:
When there are no results, we should
* show a message: »No results for **searchterm**«
* add a button »Clear search« below it
* do not show the »Load more« button

DeepDiver1975:
> Another thing: What seems to be searched is the sender and the message body. What is not searched though is the subject and people in cc. Can we add that?

we need to have a closer look - according to the api call message body and all headers are searched.

jancborchardt:
Mmmmmhh now we can search our mails

PR #476, opened 2015-04-14
reduce message list width in favor of message detail view

jancborchardt:
Instead of 35% (message list) and 65% (message detail view) width, it’s now 30% and 70%. This gives more space to the actual emails.

Please review @DeepDiver1975 @wurstchristoph


wurstchristoph:
looks good :+1:

PR #479, opened 2015-04-15
disable send/reply buttons unless recipient and either subject or message body is filled

### LINKED ISSUE #99: Disable send and reply buttons as long as message is empty
For the new message, the send button should be disabled in the beginning. It should only be enabled as soon as:
* there is at least one recipent
* and there is at least a subject OR a message body OR an attachment (in the future)

For replies, since there’s only the message body, there should at least be the message body. In the future with advanced capabilities for replies (attachments and cc/bcc), it’s probably similar to the new message logic.

jancborchardt:
Fix #99

Please review @RandolfCarter @seal20 – btw you’re also welcome in our IRC channel #owncloud-mail on freenode. :)

cc @DeepDiver1975 @wurstchristoph @zinks-

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -444,6 +444,22 @@ var Mail = {
 			$('#mail-setup').removeClass('hidden');
 			// don't show New Message button on Add account screen
 			$('#mail_new_message').hide();
+		},
+
+		toggleSendButton:function () {
+			if(($('#to').val() !== '') && (($('#subject').val() !== '') || ($('#new-message-body').val() !== ''))) {
+				$('#new-message-send').removeAttr('disabled');
+			} else {
+				$('#new-message-send').attr('disabled', true);
+			}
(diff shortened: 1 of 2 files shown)


wurstchristoph:
The button looks the same both in enables and in disabled state. If the button changes color according to its state, it would be easier to recognize by the user :-)

jancborchardt:
True, I also thought about that. Will fix it. :)

jancborchardt:
@wurstchristoph fixed it in core with https://github.com/owncloud/core/pull/15638 – please review. :)

Does this pull request work fine otherwise?

wurstchristoph:
Nice!
yes, it does! :+1:

RandolfCarter:
bothsorry for the late reply, yup, works as expected - the button is disabled until some text is entered :+1:

PR #481, opened 2015-04-15
Use to as from in sent folders - fixes #342

### LINKED ISSUE #342: Show recipient in sent-mail message list
In our 'sent mail' folder(s), the message list uses the same template as the other folders.
The recipient should be shown in the message list, not the sender (it's us!)

jancborchardt:
Please review @zinks- :)


4lexwe:
:+1: tested & works

jancborchardt:
Cool, thank you @4lexwe!

The only thing that’s missing here is the internationalization for the ` & others` – @DeepDiver1975 your turn. :)

PR #483, opened 2015-04-15
empty cc and bcc fields on send as well

jancborchardt:
Before we didn’t clear them visually like we did with the other fields. Please review @DeepDiver1975 @wurstchristoph


wurstchristoph:
nice :+1:

Issue #485, opened 2015-04-15
IMAP errors seem not to be logged.


Henni:
When sending an email the mail app tells me that there is an IMAP error.
This is done by a short message below the mail input area.
I am missing further information about this error either instead of the short message or in the log file of owncloud.

![owncloud-imap-error](https://cloud.githubusercontent.com/assets/1267827/7165967/2acb2a74-e3ab-11e4-9d25-4ff15896edfe.png)

Henni:
On a second thought the error message should definitely occur where it currently is but with more details.

It shouldn't appear (only) in the log. There shouldn't be an admin required to fix a simple mail setup error.

Xenopathic:
From what I remember there is nothing much we can do in such a situation - the IMAP server literally just replies with `BAD`. It might be a good idea to log exactly what command caused that though... Anyhow, moving to 0.2

DeepDiver1975:
since #522 the while IMAP conversation is written to data/horde.log in case debug mode is enabled in the config.php - maybe this helps ....

```php
<?php
define('DEBUG', 1);

ChristophWurst:
closing this as I don't see any (easy) way to fix this.

PR #486, opened 2015-04-15
log why sending a new mail failed

wurstchristoph:
in favor of #485

### DIFF
--- a/lib/controller/accountscontroller.php
+++ b/lib/controller/accountscontroller.php
@@ -279,6 +279,7 @@ class AccountsController extends Controller
 			$raw = stream_get_contents($mail->getRaw());
 			$sentFolder->saveMessage($raw);
 		} catch (\Horde_Exception $ex) {
+			$this->logger->error('Sending mail failed: ' . $ex->getMessage());
 			return new JSONResponse(
 				array('message' => $ex->getMessage()),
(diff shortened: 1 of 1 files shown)


Xenopathic:
It'd be a good idea to log the mail ID to aid debugging

wurstchristoph:
Do you mean something like this:

	if ($mailbox) {
		$this->logger->error('Sending a new mail failed: '
			. $ex->getMessage());
	} else {
		$this->logger->error('Sending a reply of message '
			. $messageId . ' failed: ' . $ex->getMessage());
	}

As it could be newly created mail, it doesn't necessarily have an ID at that point. Therefore an ID could only be logged if the mail is a reply

jancborchardt:
@Henni can you test this?

Xenopathic:
@wurstchristoph Hmm, I see how that is an issue. Ideally we need some way to identify which message caused the error, but hopefully in most cases the exception message will state everything that is needed (like 'failed to connect to mail.domain.com'). Leave it as it is for now :smile:

Oh, :+1: too

Henni:
I checked out this pull request and it seems that it doesn't fix my problem. (#485)
There is a message in the log but it doesn't contain any additional info.
> Sending mail failed: IMAP error reported by server.

I also commented on #485 to specify my problem in more detail.

Issue #461, opened 2015-04-11
Current folder highlight is removed when opening another unread mail


jancborchardt:
1. Go to your inbox, wait until the current message is loaded
2. Click on another _unread_ message
3. Observe how the highlight of the Inbox entry in the sidebar disappears

@DeepDiver1975 @wurstchristoph any idea?

wurstchristoph:
The following line seems to cause it: https://github.com/owncloud/mail/blob/master/js/views/folder.js#L81

Therefore I guess it's about handlebars rendering the folder element again, so the ```active``` class is removed

PR #489, opened 2015-04-15
highlight only one folder at a time

wurstchristoph:
this should finally fix https://github.com/owncloud/mail/issues/440#issuecomment-93549113
@jancborchardt


jancborchardt:
Yup, nice! :+1: 

I guess https://github.com/owncloud/mail/issues/461 is a different part, right? :)

wurstchristoph:
Exactly, #461 is caused by handlebars view rendering as far as I understood

PR #494, opened 2015-04-17
fix send/reply button being disabled again after sending a message

jancborchardt:
@wurstchristoph fixes the issue you mentioned in https://github.com/owncloud/mail/issues/99#issuecomment-93816187


wurstchristoph:
yes, it's fixed now ;-)

jancborchardt:
Thanks for the thorough testing :)

PR #495, opened 2015-04-17
use correct capitalization for 'Re:' prefix, prevent stacking

jancborchardt:
Changing from the strange-looking capitalized »RE:« prefix to the more commonly used (and less screamy/technical ;) »Re:«.

Please review @DeepDiver1975 @wurstchristoph @zinks-

### DIFF
--- a/lib/controller/accountscontroller.php
+++ b/lib/controller/accountscontroller.php
@@ -60,7 +60,7 @@ class AccountsController extends Controller
 	 * @var \OCP\Files\Folder
 	 */
 	private $userFolder;
-	
+
 	/**
 	 *
 	 * @var Logger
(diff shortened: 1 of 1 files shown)


jancborchardt:
And as bonus I added prevention of stacking multiple »Re: Re:« prefixes before a subject the longer the thread gets.

lib/controller/accountscontroller.php
```diff
@@ -231,7 +231,12 @@ public function send($accountId) {
 			$message = $mailbox->getMessage($messageId);
 
 			if (is_null($subject)) {
-				$headers['Subject'] = "RE: " . $message->getSubject();
+				// prevent 'Re: Re:' stacking
+				if(substr($message->getSubject(), 0, 4) === 'Re: ') {
```
DeepDiver1975: You might want to compare case insensitive here to catch usages of RE as well - http://php.net/manual/en/function.strcasecmp.php

jancborchardt: Thanks! Thought about it but forgot for some reason.

PR #496, opened 2015-04-17
Remove shading of date

### LINKED ISSUE #201: Date of older emails is almost unreadable
There color of the date the email was received as displayed in the message list vary from black (brand new mail) to a very light gray (last year's mail).
The date for mails more than a few months old can be very difficult to read :

screen shot 2014-08-18 at 12 45 08

People using a cheaper/mistuned monitor might not even see it.

jancborchardt:
It’s useful in the Files app because there files with different dates are directly among each other.

Mails however are always sorted by date so shading the date is not useful here. Removing it also fixes the visibility issues, fix #201

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -556,17 +556,17 @@ $(document).ready(function () {
 			}
 		});
 	});
-	
+
 	// set standard port for the selected IMAP & SMTP security
-	
+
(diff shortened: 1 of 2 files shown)


jancborchardt:
Tested by @wurstchristoph via IRC. ;)

Issue #480, opened 2015-04-15
Client-side cache of accounts/folders and message lists of already loaded folders


jancborchardt:
A simple thing to improve the perceived performance would be to cache the following things:
* [ ] accounts in the sidebar
* [x] folder list in the sidebar (which will change very rarely) PR at #766
* [x] the first 20 messages of folders which have already been loaded – most of the time we will have to append new mails up top
* [x] maybe the topmost mail in that folder which is displayed automatically

We could simply do that by saving it in the local browser database once it is loaded. Then on refresh or reload, display that data until/if the new stuff comes in.

This will improve the perceived performance quite a bit and I think we should do this for 0.1 – cc @DeepDiver1975 @wurstchristoph

jancborchardt:
First step: Account and folder caching done via https://github.com/owncloud/mail/pull/501 :)

wurstchristoph:
I would suggest we cache the 20 most recently loaded messages. I had a look at my inbox, most of the mails are under 1KB, some are up to 4KB: ~3KB average.
If I assume most users have 1 to 3 mail accounts with 5 - 15 folders each, the worst scenario would cache 20 * 3 * 15 * 3KB = 2700KB = 2.6MB.
I found out that modern browsers local storage size limit is about 5 to 10MB, which should suffice.

The great thing about caching messages in the client is that they are immutable, no further update functionality is needed, only some kind of cleanup of old cached messages.

jancborchardt:
Nice! That limit is on localStorage only though, right? IndexedDB and the newer local browser storages can handle more from what I remember.

wurstchristoph:
I'm not sure about that https://stackoverflow.com/questions/10988569/what-are-the-storage-limits-for-the-indexed-db-on-googles-chrome-browser (Although that article may be outdated)

jancborchardt:
Basically when #766 is reviewed and merged, this is fixed, right @wurstchristoph? :)

wurstchristoph:
@jancborchardt More or less, it takes still some time to load all accounts first as they are not cached

jancborchardt:
Moving the remaining part of caching accounts in the sidebar to 0.3

ChristophWurst:
Moving to 0.4 as account caching is not going to be included in 0.3

ChristophWurst:
[...]

PR #501, opened 2015-04-17
Caching accounds and folders

DeepDiver1975:
refs #4 

@jancborchardt

### DIFF
--- a/js/jquery.storageapi.js
+++ b/js/jquery.storageapi.js
@@ -1,0 +1,451 @@
+/*
+ * jQuery Storage API Plugin
+ *
+ * Copyright (c) 2013 Julien Maurel
+ *
+ * Licensed under the MIT license:
+ * http://www.opensource.org/licenses/mit-license.php
+ *
+ * Project home:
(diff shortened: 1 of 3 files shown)


jancborchardt:
Ohhh nice it’s starting. ;)

Nothing loading though. JS error:
```
TypeError: $.initNamespaceStorage is not a function mail.js:36:16
```
owncloud.log:
```
Could not find resource file \"js\/jquery.storageapi.js\/\/owncloud\/apps\/mail\"
\/owncloud\/index.php\/apps\/mail\/"
```

jancborchardt:
Also correct ref: https://github.com/owncloud/mail/issues/480

DeepDiver1975:
@jancborchardt give it another try - thx

jancborchardt:
Reaally really nice! Works pretty perfectly. :+1: 

If we now cache the message list and the first message (and maybe a few more) we’ll be golden regarding perceived performance. :)

DeepDiver1975:
> If we now cache the message list and the first message (and maybe a few more) we’ll be golden regarding perceived performance. :)

comming ..... - but not today ....

PR #502, opened 2015-04-17
Fix code checker

DeepDiver1975:


Xenopathic:
:+1: Code looks good

jancborchardt:
And stuff still works :+1:

Xenopathic:
This has broken compatibility with oC 8.0... will oC Mail 0.1 be restricted to 8.1 and above only?

jancborchardt:
Oh lala – @DeepDiver1975 we want to preserve 8.0 compatibility, right? Can you fix that?

(I’m fine with breaking ownCloud 7 compat, but 8.0 we should definitely work with.)

jancborchardt:
And @Xenopathic thanks for finding it out! We mostly test with our master installations.

Xenopathic:
@jancborchardt Hehe, I run it on my 'production' server (3 users :smile:) so I'm running the latest stable.

PR #503, opened 2015-04-17
Security mode fixes and improvements

Xenopathic:
Three small changes here:

 - SMTP default port has been changed to 587, which is the SMTP submission port, and is more suitable than port 25 which is the MTA port.
 - Security mode dropdown uses nicer labels: `None`, `SSL/TLS` and `STARTTLS`, like how Mozilla Thunderbird does it.
 - I also noticed that for no security, one must pass `false` into the Horde libraries, not `none` as we are at the moment. The third commit fixes that (and implements support for adding custom security modes in the future).

Request to be merged for 0.1 @jancborchardt ?

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -565,10 +565,14 @@ $(document).ready(function () {
 		var imapDefaultPort = 143;
 		var imapDefaultSecurePort = 993;
 
-		if ($(this).val() === 'none') {
-			$('#mail-imap-port').val(imapDefaultPort);
-		} else {
-			$('#mail-imap-port').val(imapDefaultSecurePort);
+		switch ($(this).val()) {
+			case 'none':
+			case 'tls':
+				$('#mail-imap-port').val(imapDefaultPort);
+				break;
+			case 'ssl':
+				$('#mail-imap-port').val(imapDefaultSecurePort);
+				break;
 		}
 	});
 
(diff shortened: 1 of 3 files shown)


jancborchardt:
@Xenopathic nice! Some details:
* the port for »None« and »Starttls« is the same. Is that correct?
* »None« is the default – should we not change that to »SSL/TLS«?
* and then sort the list by security first: »SSL/TLS« then »STARTTLS«, then »None«?

cc @LukasReschke @wurstchristoph @DeepDiver1975 @mikelupe for review

Xenopathic:
@jancborchardt Yes, None and STARTTLS now use the same port. This is because STARTTLS negotiates TLS over an unencrypted channel, which runs on the default port.

I haven't changed the order of the options, but if we want to move away from None as the default I'd suggest STARTTLS (it seems to be more recommended than SSL/TLS among mail providers). But TBH, this is the advanced configuration we are talking about, and a default-off option makes sense to me :smile: Also, in this order the drop down is identical to the Thunderbird configuration, albeit missing an Autodetect feature.

jancborchardt:
Ok, if it’s the same as Thunderbird I guess it’s ok. :)

@LukasReschke security review plz

clementhk:
no real logic changes here so I think it's fine in security wise... it only pre-selected the more secure method..

I don't like it having multiple return types in functions being a downvote for convertSslMode(), but I'm guessing there's not much we could do here(?)
I'd use nullable string(have you tried?)

I can confirm that none is not supported from a quick search on https://github.com/horde/horde/blob/master/framework/Imap_Client/lib/Horde/Imap/Client/Socket.php

Another thing is, don't we need to put `selected="selected"` or something on `mail-smtp-sslmode` to `STARTTLS`? Or maybe we already have some kind of js mapping? If we don't then maybe we can push this further to autofill `mail-smtp-port`.

PR #504, opened 2015-04-18
if manual setup is open, close it after successful connection, fix #331

### LINKED ISSUE #331: Failed advanced account creation breaks interface

jancborchardt:
(Manual rebase of https://github.com/owncloud/mail/pull/333)

This moves the manual setup toggle code into a function. Also fixes #331 in that the manual setup area is only closed on successful connection.

Please test @Xenopathic @zinks- @PoPoutdoor @wurstchristoph

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -497,6 +497,21 @@ var Mail = {
 			} else {
 				$('.reply-message-send').attr('disabled', true);
 			}
+		},
+
+		toggleManualSetup:function() {
+			$('#mail-setup-manual').slideToggle();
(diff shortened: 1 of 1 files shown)


Xenopathic:
Tested, works :+1:

PR #505, opened 2015-04-18
save and open drafts

### LINKED ISSUE #117: Save messages in progress as draft
Messages you are writing (replies as well) should be saved as drafts. This needs to happen via autosave instead of a dedicated »save« button. Every few seconds it should be saved.

If that’s too much for the mailserver, ideally save it to ownCloud like 500ms after typing is stopped. That’s how we do it in Notes I think.

wurstchristoph:
# Current progress
## IMAP drafts
- [x] open/edit drafts
- [x] store new message drafts
- [ ] store reply message drafts

## Local storage drafts
- [ ] store new message drafts
- [x] store reply message drafts

## Issues to fix
- [ ] ~~attachments are ignored~~ will be fixed in a future PR :smile_cat: 
- [x] HTML drafts do not work properly
- [x] when opening an existing draft, the message id (IMAP uid) needs to be updated if the draft is updated since it get's a new UID from the IMAP server. This unfortunately breaks the message list.
- [x] delete draft after sending it
- [ ] ~~flag replied message as reply when sending the draft~~ not possible AFAIK
- [ ] toggle cc when loading reply drafts if it is set
- [x] merge https://github.com/owncloud/mail/pull/545

fixes #117

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -13,6 +13,7 @@ $app->registerRoutes($this,
 			['name' => 'page#index', 'url' => '/', 'verb' => 'GET'],
 			['name' => 'page#compose', 'url' => '/compose', 'verb' => 'GET'],
 			['name' => 'accounts#send', 'url' => '/accounts/{accountId}/send', 'verb' => 'POST'],
+			['name' => 'accounts#draft', 'url' => '/accounts/{accountId}/draft', 'verb' => 'POST'],
 			['name' => 'accounts#autoComplete', 'url' => '/accounts/autoComplete', 'verb' => 'GET'],
 			[
 				'name' => 'messages#downloadAttachment',
--- a/js/mail.js
+++ b/js/mail.js
@@ -498,6 +498,10 @@ var Mail = {
 				Mail.State.composeView.data = data;
 			}
 
+			if (data && data.hasHtmlBody) {
+				Mail.UI.showError(t('mail', 'Opening HTML drafts is not supported yet.'));
+			}
+
 			Mail.State.composeView.attachments.reset();
 			Mail.State.composeView.render();
 
--- a/js/mail.js
+++ b/js/mail.js
@@ -43,13 +43,13 @@ var Mail = {
 					icon: icon
 				}
 			);
-			notification.onclick = function() {
+			notification.onclick = function(x) {
 				Mail.UI.loadMessages(accountId, folderId, false);
 				window.focus();
 			};
 			setTimeout(function () {
 				notification.close();
-			}, 10000);
+			}, 5000);
 		}, checkForNotifications: function() {
 			_.each(Mail.State.accounts, function (a) {
 				var localAccount = Mail.State.folderView.collection.get(a.accountId);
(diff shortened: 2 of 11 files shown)


DeepDiver1975:
@wurstchristoph WOW - thanks a lot for implementing this! Looking forward to test this!

jancborchardt:
@wurstchristoph nice, saving it already works very nice! Just one problem: When you click »Save Draft« the email also gets sent. ;)

wurstchristoph:
@jancborchardt thanks for the hint! 
Removing lines https://github.com/wurstchristoph/mail/blob/drafts/lib/controller/accountscontroller.php#L356-L357
prevents drafts being sent. This, however, introduces another issue:

    {"reqId":"1601b1157ecca7fa80588a768f4f898a","remoteAddr":"::1","app":"mail","message":"Saving draft failed: No base part set.","level":3,"time":"2015-04-19T21:12:06+00:00","method":"POST","url":"\/index.php\/apps\/mail\/accounts\/10\/draft"}

The problem here is: The mail is not sent, so ```$mail->getRaw()``` fails due to no base part being available. I haven't found any resource describing how to append ```Horde_Mime_Mail``` to a folder.

@DeepDiver1975 can you help with that?

Unfortunately I haven't found more documentation of the horde library than this: https://github.com/horde/horde/blob/master/framework/Imap_Client/doc/Horde/Imap/Client/examples/tutorial.md

wurstchristoph:
Should attachments be saved in drafts too?
If not, ```append``` can be used easily like in the horde tutorial (link posted above). Moreover I have no idea on how to resume drafts that already have attachments. Additionally, if we want to store drafts also in the browser locally, there is probably no easy way to do this (thinking of uploaded files we will hopefully support in future).
Summing up, I think I'm capable of implementing storing and resuming text-only mails. If we want to safe attachments too, unfortunately this task seems to be too complicated for me :/

@DeepDiver1975 @jancborchardt what's your opinion?

jancborchardt:
@wurstchristoph it’s all a matter of iteration. ;) So it’s totally OK if the first version can only save text-only drafts. :)

wurstchristoph:
Open/Store/Edit of simple text messages works more or less (there are still issues with HTML messages)
Since [IMAP messages are immutable](http://stackoverflow.com/questions/16945041/imap-editing-a-draft-or-existing-message), updating drafts is done by creating a new draft an deleting the old one. This means the draft will get a new UID. This breaks the message (draft) list: you cannot click on an updated draft again.

I'll now have a look at automatically storing drafts... :smile:

jancborchardt:
@wurstchristoph best rebase this since the password encryption was merged to master – #510 :)

wurstchristoph:
@jancborchardt done ;-)

Xenopathic:
@wurstchristoph Needs a rebase

wurstchristoph:
@Xenopathic rebased.

I had many conflicts that had to be resolved manually. Hopefully I didn't merge something away :worried:

jancborchardt:
@wurstchristoph it's about time we get this done, tested and merged! :) We can do it in separate pull requests, just step by step.

wurstchristoph:
Alright, saving and resuming drafts should now work as expected. Reply messages are only stored to local storage for now, I propose to implement this in a future PR as it requires some rewrite/DRY-out of different composers (new-message/reply).

HTML is a problem for know, as it cannot simply be loaded into the textarea. As discussed with @DeepDiver1975 on IRC, we'll stay with simple text messages for now. HTML handling can be added in future.

There is only one minor issues left: cc/bcc toggle should be opened by default when resuming a draft message where one of those fields isn't empty. I haven't found a way to accomplish that unfortunately :/

If there isn't something crucial I should add in this PR, I propose we change the state to 'review'.

Could you please test/review? @DeepDiver1975 @jancborchardt @Xenopathic @zinks- @LukasReschke

Xenopathic:
Why are there there many commits with duplicate names? Could you squash some of them down?

wurstchristoph:
@Xenopathic I noticed them too, but could not figure out how I created them :/

I will squash the commits anyway when I'm done implementing ;-)

jancborchardt:
@wurstchristoph nice! If you can not figure out expanding the cc/bcc stuff, just leave it – we can fix that later.

Just squash your commits, set this to review and let us know. Then we can review, test and merge this! :)

wurstchristoph:
rebased and squashed.

@jancborchardt please set it to review, I don't have the permission to change that ;-)

jancborchardt:
@wurstchristoph niiice! Works very well. :) Just a few details:
Drafts have a `<br/>` appended at every line (including the last one) even though the mails are not HTML:
```
Testing testing<br />
Drafts<br />
<br />
```

* when continuing a draft, the »New message« button is deactivated and the highlight of that draft in the message list is lost. I assume that’s because of the hijacking of the »New message« form. But it would be good if both highlight and just starting a new message would still work.

Those are the only bugs I could find. Can you fix those @wurstchristoph? If you can’t fix the highlight/newmessage one that’s ok. But the `<br/>` insertion bug is the one which should really be fixed.

(What was the reason again for »store new message drafts in localStorage« not working yet? Cause when that would be in that’d be pretty nice. To just be able to navigate back to the »New message« and the latest state would be preserved.)


TL;DR ONLY FIX THE `<br/>` BUG AND DON'T LISTEN TO MY OTHER FEATURE REQUESTS SO WE CAN MERGE THIS :D (ALSO YOU’RE AWESOME)

jancborchardt:
Oooh and each time opening the draft the encoding gets crazier. 2nd time:
```
Testing testing&lt;br /&gt;<br />
Drafts&lt;br /&gt;<br />
&lt;br /&gt;<br />
<br />
```
3rd time:
```
Testing testing&amp;lt;br /&amp;gt;&lt;br /&gt;<br />
Drafts&amp;lt;br /&amp;gt;&lt;br /&gt;<br />
&amp;lt;br /&amp;gt;&lt;br /&gt;<br />
&lt;br /&gt;<br />
<br />
```

You get the idea. ;)

wurstchristoph:
I know about this bug ;-) That's what we talked about on IRC yesterday.

If there's some time I'll have a look at it later, I have to figure out who is creating those line breaks and where the message is converted to HTML. Both things have to be removed.

zinks-:
Very nice, I'm loving this feature! :+1:

jancborchardt:
@wurstchristoph aaaand merged! :) Can you open a follow-up issue for the remaining issues for 0.2?

jancborchardt:
@wurstchristoph just wanted to let you know that drafts are damn awesome! :rocket: great work! :)

wurstchristoph:
Thank you @jancborchardt :smiley:

PR #514, opened 2015-04-20
Don't load multiple folders at once

### LINKED ISSUE #467: Multiple folders get loaded when switching fast
As @wurstchristoph mentioned in issue #440:

>  if you switch folders that fast, the message column is loaded for each folder you clicked (ajax responses coming in). This looks odd and should be fixed too

When another folder is clicked, the previous ajax call should probably be cancelled right away.

@DeepDiver1975 @zinks- something for you? :)

zinks-:
Reference to the currently loading request is kept in Mail.State and aborted if another folder is loaded.
TextStatus from the $.ajax error handler is trickled down to prevent handling it as a real error.

EDIT: changed the PR name to something more explicit ;)

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -4,6 +4,7 @@ var Mail = {
 		currentFolderId: null,
 		currentAccountId: null,
 		currentMessageId: null,
+		currentlyLoading: null,
 		accounts: null,
 		messageView: null,
 		router: null,
@@ -44,11 +45,11 @@ var Mail = {
 					return;
 				}
 			}
-			$.ajax(url, {
+			return $.ajax(url, {
 				data: {},
 				type: 'GET',
-				error: function () {
-					options.error();
+				error: function (xhr, textStatus, errorThrown) {
+					options.error(textStatus);
 				},
 				success: function (data) {
 					if (allOptions.cache) {
@@ -238,6 +239,9 @@ var Mail = {
 		},
 
 		loadMessages: function (accountId, folderId, noSelect) {
+			if (Mail.State.currentlyLoading !== null) {
+			    Mail.State.currentlyLoading.abort();
+			}
 			// Set folder active
 			Mail.UI.setFolderActive(accountId, folderId);
 			Mail.UI.clearMessages();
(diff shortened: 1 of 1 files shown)


zinks-:
@jancborchardt @DeepDiver1975 @wurstchristoph Please review

DeepDiver1975:
unit tests on travis will fail because the encrypted test credentials cannot be decrypted

js/mail.js
```diff
@@ -4,6 +4,7 @@ var Mail = {
 		currentFolderId: null,
 		currentAccountId: null,
 		currentMessageId: null,
+                currentlyLoading: null,
```
wurstchristoph: you used spaces here, they should be replaced by tabs ;-)

jancborchardt:
Otherwise it looks like it’s fixed for me. :+1: 

@DeepDiver1975 would rebasing fix the unit tests?

js/mail.js
zinks-: We should have a .editorconfig :D

jancborchardt: So, can you fix it? :) Then we can merge it I guess. @wurstchristoph did you test it and also works for you?

jancborchardt:
@zinks- can you replace the spaces with tabs please? ;)

js/mail.js
wurstchristoph: Yes, i did and it works :+1:

DeepDiver1975:
> @DeepDiver1975 would rebasing fix the unit tests?

no - the "issue" is that this PR is coming from a different repo - the secured data can then not be read by travis because the private key is within our repo - this is explained somewhere in the travis docs ....

zinks-:
Messed up my commit history, but the result should me mergeable (had a merge conflict) and has only tabs for indent.
If you can tell me what git commands to use to clean the history up I will gladly do it.

jancborchardt:
@zinks- something with "squash". That allows you to put all the changes you made into one commit.

zinks-:
@jancborchardt thanks, it's done.

jancborchardt:
@zinks- nice! Merged. :)

PR #518, opened 2015-04-20
Set type of port inputs to "number" - fixes #517

Henni:

PR #521, opened 2015-04-21
Set type of user inputs to text instead of email - fixes #520

### LINKED ISSUE #520: Inputs for IMAP/SMTP user shouldn't necessarily be an email address
The input fields for the IMAP and SMTP users are set to type="email". But it should also be possible to set the username to a simple string instead of an email address. Therefore the type should be replaced by "text".
( similar to #517 )

Henni:
fixes #520

### DIFF
--- a/templates/index.php
+++ b/templates/index.php
@@ -310,7 +310,7 @@ script('mail', 'jquery.storageapi');
 					<label for="mail-imap-port" class="infield"><?php p($l->t('IMAP Port')); ?></label>
 				</p>
 				<p class="groupmiddle">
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
Looks good - thx

jancborchardt:
@Henni even better: Set it to `number`. :)

jancborchardt:
Ah nevermind – I misread the title and thought this was https://github.com/owncloud/mail/pull/518 :D

jancborchardt:
:+1:

Issue #393, opened 2014-10-16
Use html5 notifications to inform the user about new messages.


DeepDiver1975:
https://developer.mozilla.org/en-US/demos/detail/html5-notifications

jancborchardt:
Setting to 0.3 for now as it’s a nice-to-have but not essential.

PR #522, opened 2015-04-21
Introduces background checking of mailboxes based on uidvalidity and uid...

DeepDiver1975:
Fixes https://github.com/owncloud/mail/issues/393

@jancborchardt we need a nice icon for html5 notification

Please test and let me know what you think!

TODO:

* [ ] fix Notification not showing in Firefox
* [ ] fix sent mails being marked as unread #539
- [x] ~~JS: update localstorge for folders~~ #536
- [x] IMAP: only query for new messages where UID > last uidnext
- [x] JS: new messages need to be inserted at the right location
- [x] JS: update browser title with unread count of INBOX
* [x] UI: show account email instead of »ownCloud Mail« in notification title
* [x] UI: add timeout of 5 seconds to the notification
* [x] UI: change favicon to favicon-notification when there are new emails – change it back once the window was opened
* [x] UI: if possible, add event handler so that clicking the notification opens the app
* [x] ~~IMAP: use mailbox status on 'Checking messages ...'~~ https://github.com/owncloud/mail/issues/538


clementhk:
Cool!
Only problem I see is multi-language support.

js/mail.js
```diff
@@ -22,6 +22,52 @@ var Mail = {
...
+						var folders = localAccount.get('folders');
+						_.each(jsondata.folders, function(f) {
+							var localFolder = folders.get(f.id);
+							if (f.uidvalidity !== localFolder.get('uidvalidity') ||
+								f.uidnext !== localFolder.get('uidnext')) {
+
+								// send notification
+								Notification.requestPermission(function() {
+									new Notification(
+										"New Mail arrived!",
```
jancborchardt: Can we here directly display sender and subject of the email? First line sender, second line subject, just like the message list.

We need to keep the notification short and sweet. »New mail« is evident from the notification itself. ;)

If there are multiple new emails, we could show it a bit different like:
```
5 new emails
from Thomas Müller, Christoph Wurst, …
```

DeepDiver1975: With the current implementation we only know that there was a change on a folder.
And as a result we can reload the current selected folder and update the unread mail counter on all folders.

To get an idea on what is new we'd need to use a different approach .... but doable.

DeepDiver1975:
> Only problem I see is multi-language support.

Can you elaborate on your feedback? You mean that the language of the notification is defined by owncloud and does not respect the desktop language?

jancborchardt:
How can I exactly test that? Send a new email to one of the accounts, leave the tab in the background and just wait?

Neither Chrome nor Firefox ask me if they can send me notifications. Shouldn’t they do that for HTML5 Notifications?

jancborchardt:
Huh, Chrome now asked me, but only after the first new email came in. Shouldn’t it ask that right away?

jancborchardt:
Ok, just got an influx of notifications about all folders even though there were no new messages:
notifications
There were more when clicking them away. Not sure which browser this was from because I had both open.

I’ll fix the icon and the wording right now. :)

jancborchardt:
@DeepDiver1975 fixed icon and wording. Is it not possible to use SVG there?

jancborchardt:
And when clicking a notification, it should probably switch to the browser, to that tab, and open the folder, right?

DeepDiver1975:
> Is that possible?

yes

DeepDiver1975:
> Is it not possible to use SVG there?

lmgtfy ...

DeepDiver1975:
> Is it not possible to use SVG there?

The icon's size should be 32x32px, supported formats: jpg/png/ico} - see https://github.com/ttsvetko/HTML5-Desktop-Notifications#browsers-support

jancborchardt:
@DeepDiver1975 kk, the new icon is 32*32 as well.

DeepDiver1975:
@jancborchardt 

> If there are multiple new emails, we could show it a bit different like:

bildschirmfoto von 2015-04-23 00 03 47

jancborchardt:
This looks awesome! If it’s from more people than just Thomas Müller can we show the first two like this:
> Thomas Müller, Christoph Wurst, …

DeepDiver1975:
> If it’s from more people than just Thomas Müller can we show the first two like this:

sure - on it ....

DeepDiver1975:
bildschirmfoto von 2015-04-23 00 24 50

jancborchardt:
Nice! Needs a rebase now ;)

DeepDiver1975:
> Nice! Needs a rebase now ;)

done

DeepDiver1975:
@jancborchardt shall we add the number of unread messages to the browser title?

Changing the number will result in that glow effect (at least on FF) which gives some feedback to the user as well.

jancborchardt:
Yup, add it to the front of the title in parentheses, like `(3) Mail - ownCloud`.

Apps can also define their own favicons, right? Then we can use a Mail icon instead of the standard one. And I can also add an icon with a red dot to indicate new messages.

DeepDiver1975:
> Apps can also define their own favicons, right? Then we can use a Mail icon instead of the standard one. And I can also add an icon with a red dot to indicate new messages.

gimme da icons - I'll add them

DeepDiver1975:
> Yup, add it to the front of the title in parentheses, like (3) Mail - ownCloud.

is that the common convention? I think I remember the counter is at the end????

jancborchardt:
Yeah, it should be in the beginning so as to be visible when you have more than 5 tabs open – that’s the use. ;)

Will do the icons now.

DeepDiver1975:
Furthermore: which total number are we going to display?

Sum of all unread in all inboxes?

jancborchardt:
@DeepDiver1975 yup, sum of unread in all inboxes sounds good.

Icons added. Not sure if the favicon-touch icons are needed at all, but I added them.

jancborchardt:
I still get notifications for »new« mail in Trash, Sent, Favorites etc. – where by definition there can be no new mail. It should only check for actually unread messages, and only new ones which came in since you last looked. And special folders like Trash, Junk, Sent, Archive and Favorites can automatically be ignored.

jancborchardt:
And another special case would be cool if there is just one single new email. In that case in the first line we should show the sender, and in the second line the first x chars of the subject (let’s say 25 characters because that seems to fit in the notification).

jancborchardt:
Here’s how to do the click handler: https://stackoverflow.com/questions/4906976/how-to-get-focus-to-a-chrome-tab-which-created-desktop-notification (TL;DR: `window.focus();`)

Doesn’t seem to work in Firefox though: https://stackoverflow.com/questions/21016999/how-to-get-focus-to-the-tab-when-a-desktop-notification-is-clicked-in-firefox

Xenopathic:
Being fixed in Firefox it seems: https://bugzilla.mozilla.org/show_bug.cgi?id=874050

Oh, and it's in the standard for this to work too: https://github.com/whatwg/notifications/commit/4b8ec1b3a7b212229e5d2b42edd5dfdb90b8401d

jancborchardt:
Rebase please cause the password encryption was merged. ;)

jancborchardt:
Aand please give it a final round of testing @wurstchristoph @Xenopathic @zinks- everyone :)

jancborchardt:
@DeepDiver1975 btw at least in Chrome for me the notification click to switch to the browser and the correct tab works. The only thing missing is to actually load the correct folder.

DeepDiver1975:
> The only thing missing is to actually load the correct folder.

Please test @jancborchardt

jancborchardt:
@DeepDiver1975 only highlights the folder, but doesn’t actually load it.

jancborchardt:
Cool, works now. 

(Btw: In Chrome it doesn’t work after all switching from another app. It only worked when I was in a different browser window. But IRCCloud also doesn’t manage that so I guess that’s fine.)

jancborchardt:
@wurstchristoph @zinks- @Gomez @clementhk @Henni @irgendwie this is ready to test now! :) Accept notifications in your browser and check this out. Every 5 minutes the app will check for new mails. Works as long as ownCloud Mail is open in some tab in the browser.

jancborchardt:
@DeepDiver1975 maybe good to rebase once more to have the disabled cache in here as well.

DeepDiver1975:
@jancborchardt rebased and *hopefully* travis fixed for now - needs more love - but this can be done in a separate PR.

js/mail.js
```diff
@@ -66,6 +151,28 @@ var Mail = {
 		}
 	},
 	UI: {
+		changeFavicon: function (src) {
+			$("link[rel='shortcut icon'").attr("href",src);
```
jancborchardt: The notifications and check doesn’t work in Firefox seemingly due to this selector:
`Error: Syntax error, unrecognized expression: link[rel='shortcut icon' jquery.min.js:4:15709`

@DeepDiver1975

jancborchardt: Heh – closing brace missing. Will fix and test.

jancborchardt:
@DeepDiver1975 I fixed the syntax error which caused problems in Firefox, but there it still doesn’t quite seem to work.

jancborchardt:
@DeepDiver1975 ok, added the two last things I found to the checkbox list in the top.

clementhk:
Wow it looks a lot better than the initial version from screenshots, didn't know this would have so much focus, nice work guys.

clementhk:
I think you guys also noticed Travis failed.

clementhk:
Since it only fails on specific versions, one thinking is that the tests ran parallelized across environments, meaning php 5.4 test deleted/created mailboxes, and when php 5.5/5.6's tests tries to do the same/related things, then tests failed because there are linked towards the same account, but this is just one guess.

Also I noticed you might need "composer self-update" in your travis? At least it is complaining that.

jancborchardt:
I also just got a notification for »0 new mails in Iinbox«. ;)
Probably a glitch, but we should catch that.

jancborchardt:
We should get this merged before it gets too big. :O

@DeepDiver1975 can you fix the "sent messages marked as unread" issue caused here? Everything else we can fix after merge.

DeepDiver1975:
> We should get this merged before it gets too big. :O

yes - needs to be merged finally - #539 should be fixed now as well

jancborchardt:
Ok, works nicely! Merging :) :+1:

PR #531, opened 2015-04-23
[WIP] cache message list and messages locally

wurstchristoph:
As described in #480 it is a great performance boost to cache messages locally.
Since IMAP messages have an unique UID, messages can be cached without ever needing to update them.

# Current progress
- [ ] cache message list
- [x] cache messages
- [ ] get rid of old messages (ttl)

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -1,4 +1,4 @@
-/* global Handlebars, Marionette, Notification, relative_modified_date, formatDate, humanFileSize, views */
+/* global Handlebars, Marionette, relative_modified_date, formatDate, humanFileSize, views */
 var Mail = {
 	State: {
 		currentFolderId: null,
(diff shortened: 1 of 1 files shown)


jancborchardt:
Whoo, great stuff! :) Let me know as soon as the message list is cached as well and I’ll test.

jancborchardt:
@wurstchristoph so we’re actually thinking about reverting the caching for the accounts/folders list  #501 because it caused so many issues.

Maybe better something for 0.2? What do you think @DeepDiver1975 @wurstchristoph @zinks- @Gomez @Xenopathic @PoPoutdoor everyone?

Xenopathic:
Looking at the facts:

* Caching breaks deleting accounts, and is reported to break on creation of accounts.
* oC Mail 0.1 is due to be released in 11 days time
* The issue report has been around for a while, and no obvious fix has been found

I vote for reverting until after 0.1

jancborchardt:
Disabled caching for now via https://github.com/owncloud/mail/pull/536

wurstchristoph:
I also think it's a good idea to move this to 0.2

jancborchardt:
@wurstchristoph you mentioned on IRC you feel this is outdated and want to close it. That’s fine – just go ahead as we have https://github.com/owncloud/mail/issues/480 to track the general feature. :) Maybe also update that with your findings and how we could proceed better,

PR #535, opened 2015-04-24
Style empty content messages

jancborchardt:
Better design for:
* initial mail account connection
* empty mail folders
* no search results

Only thing left to do: Hide `#app-navigation` and `#app-navigation-toggle` on first time setup. But only when no account is configured yet. Please help @DeepDiver1975

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -38,13 +38,18 @@ button.control {
 /* SETUP */
 #mail-setup {
 	position: absolute;
+	z-index: 1001;
 	width: 250px;
 	top: 10%;
 	left: 40%;
 	padding-bottom: 50px;
 }
 #mail-setup h2 {
-	text-align:center;
+	text-align: center;
+	position: relative;
+	width: 300%;
+	left: -100%;
+	margin-bottom: 20px;
 }
 #mail-setup input {
 	width: 100%;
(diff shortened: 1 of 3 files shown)


jancborchardt:
Ready for review @Xenopathic @zinks- @Gomez

jancborchardt:
Oi, we need a review here :) @clementhk @irgendwie @Henni and the others

Xenopathic:
When searching for something, there's a couple of things that look off.

1) The text in the search box is too close to the search icon
2) The currently loaded message remains visible, with 'No search results for foobar' visible where the message list would be. I think the currently visible message should disappear...

Also, with the 'no mails' empty folder view, the 'Check messages' button looks out of place. Perhaps it'd be better suited to the sidebar?

jancborchardt:
> 1) The text in the search box is too close to the search icon

Looks fine for me. Can you supply a screenshot? If anything though that’s a core issue.

> 2) The currently loaded message remains visible, with 'No search results for foobar' visible where the message list would be. I think the currently visible message should disappear...

Hm, not sure about that. I actually think it’s good that the current message still stays because it leaves you with some anchor. What do you think @wurstchristoph @DeepDiver1975 @irgendwie?

> Also, with the 'no mails' empty folder view, the 'Check messages' button looks out of place. Perhaps it'd be better suited to the sidebar?

I’d prefer for now if it stays where it normally is already. Hopefully it just goes away with pull to refresh #104 ;)

jancborchardt:
Merging for now, any improvements we can do later.

PR #536, opened 2015-04-24
don't cache for the time being

### LINKED ISSUE #524: App doesn't work in Firefox: Account creation screen is shown everytime the app is opened
I can't use the mail app in Firefox. I've had the case with **two different owncloud servers** (8.0.2 and daily build) + mail master.

Problem has been observed with latest Iceweasel, latest official Firefox on Debian 8 and Mac OS 10.10.3.

After I set up the first account, I can see my mail.
However If I navigate away and come back to the mail app (or reload the page), I am only shown the "create new account" page.

An ajax request to `https://server/owncloud/index.php/apps/mail/accounts` is made and seems to return the proper account details, but nothing else happens after that.

It works in Chrome (and chromium) on Debian 8 and Mac OS.

DeepDiver1975:
Fix #524

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -33,6 +33,8 @@ var Mail = {
 				allOptions = options || {};
 			_.defaults(allOptions, defaultOptions);
 
+			// don't cache for the time being
+			allOptions.cache = false;
 			if (allOptions.cache) {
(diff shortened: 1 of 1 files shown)


jancborchardt:
Please review @zinks- @Henni @ArndOtten @wurstchristoph

PR #544, opened 2015-04-27
First functional version of expunging messages.

### LINKED ISSUE #339: Deleting messages from trash
Right now there is no way to remove a message from the trash: the 'delete' icon only moves the message to the same folder (= does nothing)

We could
1. completely remove the 'delete' icons from the trashed messages and only let people empty the trash as a whole
2. or have the delete icons really delete and purge the messages
3. or both : let people delete select messages from the trash, and empty the whole trash if they want.

irgendwie:
do **not** merge yet!
fixes #339

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Mail app
 [![Code Coverage](https://scrutinizer-ci.com/g/owncloud/mail/badges/coverage.png?b=master)](https://scrutinizer-ci.com/g/owncloud/mail/?branch=master)
 [![Dependency Status](https://www.versioneye.com/user/projects/54e50fadd1ec5734f400078a/badge.svg?style=flat)](https://www.versioneye.com/user/projects/54e50fadd1ec5734f400078a)
 
-An email app for [ownCloud](https://owncloud.org) (minimum version 7). At the moment it is a basic IMAP client and in heavy development. (A possibility for the future would be to also have it be a dedicated email server. But that would always be optional, require additional dependencies and is out of scope for now.)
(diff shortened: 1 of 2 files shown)


jancborchardt:
@DeepDiver1975 @zinks- @wurstchristoph any feedback on the code at this point already? :)

irgendwie:
@jancborchardt I tried to get a different icon (than the trash icon) for mails in trash (the suggested 'X' sign), but having a hard time with that.
How can I check, if the Mail is located in the trash or not?

https://github.com/owncloud/mail/blob/master/templates/index.php#L78

jancborchardt:
@DeepDiver1975 is that info available in the template?

Xenopathic:
@irgendwie What's the progress on this? We are dangerously close to the release of 0.1...

irgendwie:
@Xenopathic It's functional, instead of moving the mail to trash it deletes the message.
Drawback is, that you **cannot undo** that action and there is no warning or anything.
I was waiting for @DeepDiver1975's Feedback, to change the Icon in the Trash from the delete icon to a red cross sign, to symbolize that.

jancborchardt:
@irgendwie I would say it’s ok for the first pass to not have an undo or warning there. Maybe what suffices is to use a tipsy tooltip which says »Delete permanently« and pops up when you hover over the delete icon.

You would just add the title via the `title` attribute, and then invoke tipsy in the JS via: `$('class of the delete function').tipsy({gravity:'e', live:true});`

jancborchardt:
And @Xenopathic @DeepDiver1975 do you have an answer to @irgendwie’s question above?

jancborchardt:
@irgendwie it seems that some of my commits also got into this. Best is probably to open a new pull request with a fresh branch from master with a single commit with your changes. :)

irgendwie:
New pull request #574

Issue #529, opened 2015-04-23
new mail is sent mutliple times


wurstchristoph:
I have experienced some odd behavior when I wanted to send a message (or save it as draft): Although the button is only clicked once, the messages is sent multiple times.
I noticed that it was dependent on how often I switched folders or clicked on 'New message'.

Steps to reproduce:
1. open the mail app
2. click 'New Message'
3. switch to some folder (e.g. INBOX)
4. repeat steps 2 and 3 a few times
5. actually send a new message
6. the message is sent multiple times

My guess would be that ```js/views/sendmail.js``` is intitalized multiple times which results in event (click) handlers being registered multiple times

Tested with FF

@jancborchardt seems to be the same issue you experienced. can you reproduce this?

jancborchardt:
Yeah, also experienced that. Can you reproduce using the steps above @irgendwie @Henni @zinks-?

zinks-:
And they keep adding up!
I clicked 4 different folders before sending "test", 2 folders before sending "test2", and only 1 before sending "test3"
![multiplemail](https://cloud.githubusercontent.com/assets/8377141/7305996/85b6ced6-ea02-11e4-8fc0-21a82a3e12d1.png)

jancborchardt:
Setting priority for this since it’s the most core functionality. cc @DeepDiver1975 let’s debug this tomorrow.

jancborchardt:
Fixed by @zinks- through https://github.com/owncloud/mail/pull/545 :) :rocket:

PR #545, opened 2015-04-27
Fix multiple sending of messages

zinks-:
Fix for #529 
Always using the same SendMail view instance seems to fix the problem.
However, I am not sure if it should be stored in Mail.State or Mail.UI ?

@jancborchardt @DeepDiver1975 @wurstchristoph and others:) Please review


DeepDiver1975:
perfect fix - THX :+1:

PR #550, opened 2015-04-28
Fix logging of message ID on move to trash

Xenopathic:
Array to string conversion error in log. We log message IDs in other places too, so this makes sense. I added the mailbox ID so the admin has some idea of the context.

### DIFF
--- a/lib/account.php
+++ b/lib/account.php
@@ -295,7 +295,8 @@ class Account {
 		$result = $this->getImapConnection()->copy($hordeSourceMailBox, $hordeTrashMailBox,
 			array('create' => $createTrash, 'move' => true, 'ids' => $hordeMessageIds));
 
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
great - thx. In addition we could json decode $result and log it as well.

Xenopathic:
Well, I just wanted to avoid the PHP warning about the conversion. If you want to json_decode() something feel free to push directly here :smile:

jancborchardt:
@Xenopathic @DeepDiver1975 so – merge or add something? ;)

Xenopathic:
I'm happy as it is - it actually does something useful now. Perhaps not as useful as it could be, but in the words of a great prophet:

> any improvements can come later.

:smile:

jancborchardt:
@Xenopathic lol!

PR #554, opened 2015-04-29
configurable option to enable server side caching

DeepDiver1975:
This will bring back server side caching - this time configurable and using memory cache mechanisms

To enable mail caching specify this in the config.php

``` php
'app.mail.server-side-cache.enabled' => true
```

Furthermore a memory cache mechanism has to be enabled in the config - e.g. in 8.1 set  

``` php
'memcache.local' => '\OC\Memcache\APCu',
```

@jancborchardt @Gomez

### DIFF
--- a/lib/account.php
+++ b/lib/account.php
@@ -15,6 +15,7 @@ use Horde_Imap_Client_Mailbox;
 use Horde_Imap_Client_Socket;
 use Horde_Imap_Client;
 use Horde_Mail_Transport_Smtphorde;
+use OCA\Mail\Cache\Cache;
 use OCA\Mail\Db\MailAccount;
 use OCP\IConfig;
 
@@ -91,6 +92,16 @@ class Account {
 			];
 			if ($this->config->getSystemValue('app.mail.imaplog.enabled', false)) {
 				$params['debug'] = $this->config->getSystemValue('datadirectory') . '/horde.log';
+			}
+			if ($this->config->getSystemValue('app.mail.server-side-cache.enabled', false)) {
+				$cacheFac = \OC::$server->getMemCacheFactory();
+				if ($cacheFac->isAvailable()) {
+					$params['cache'] = [
+						'backend' => new Cache(array(
+							'cacheob' => $cacheFac->create(md5($this->getId() . $this->getEMailAddress()))
+						))];
+				}
+
 			}
 			$this->client = new \Horde_Imap_Client_Socket($params);
 			$this->client->login();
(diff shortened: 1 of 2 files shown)


Gomez:
So this is a oC 8.1 feature only?
Is it the same for php 5.5 with OPCache?

DeepDiver1975:
> So this is a oC 8.1 feature only?

should work with 8.0 as well - requires apc or xcache or memcached

jancborchardt:
We need a bit more reviewers here @wurstchristoph @clementhk @irgendwie @Henni @zinks- @LukasReschke @Xenopathic summoning everyone :D

lib/account.php
```diff
@@ -92,6 +93,16 @@ public function getImapConnection() {
 			if ($this->config->getSystemValue('app.mail.imaplog.enabled', false)) {
 				$params['debug'] = $this->config->getSystemValue('datadirectory') . '/horde.log';
 			}
+			if ($this->config->getSystemValue('app.mail.server-side-cache.enabled', false)) {
+				$cacheFac = \OC::$server->getMemCacheFactory();
```
Xenopathic: Eww... can we not use dependency injection for this? Failing that, move this line to the constructor, where `$this->config` et al are set?

DeepDiver1975: for sure .. to be honest: I quickly hacked this together during a 2hrs management call :zzz:  :speak_no_evil:

jancborchardt: LOL

jancborchardt:
Not sure if there’s a change or if the internet here is just faster. :D at least nothing breaks, looks good to me.

Any way we can have the parameters automatically added to the config.php to enable caching by default?

jancborchardt:
@DeepDiver1975 ^ answers on my question above?

DeepDiver1975:
> Any way we can have the parameters automatically added to the config.php to enable caching by default?

well - based on the issues we had in the past with respect to enabled server side caching - I prefer to have it disabled by default.

jancborchardt:
Ok. I’d merge it then, looked ok when testing.

DeepDiver1975:
> Ok. I’d merge it then, looked ok when testing.

this should be fixed before merge: https://github.com/owncloud/mail/pull/554#discussion_r29363854

DeepDiver1975:
@Xenopathic feel free to push the necessary change in here - THX

Xenopathic:
I update README.md with the config.php parameter. Please review @Gomez !

DeepDiver1975:
THX @Xenopathic

Issue #466, opened 2015-04-13
Fix accessibility


jancborchardt:
Currently it’s not possible to tab through the sidebar nor through the message list as they are not recognized as links. We need to fix that.

jancborchardt:
Postponing to 0.3 cause it requires some heavy changes.

ChristophWurst:
> Postponing to 0.3 cause it requires some heavy changes.

How about fixing that in 0.5? :-)

PR #559, opened 2015-04-30
[WIP] improve accessibility

jancborchardt:
Adding tabindex attribute to folders and messages – now the folders and messages are tabbable with visual feedback. Ref https://github.com/owncloud/mail/issues/466

However, on pressing enter the folder or message is not opened yet. Can we hook up an event to that apart from only the click event? cc @DeepDiver1975 @wurstchristoph @Xenopathic?

### DIFF
--- a/templates/index.php
+++ b/templates/index.php
@@ -4,14 +4,13 @@ script('mail', 'jquery-visibility');
 ?>
 <script id="mail-folder-template" type="text/x-handlebars-template">
 	<li data-folder_id="{{id}}" data-no_select="{{noSelect}}"
-		class="
-		{{#if unseen}}unread{{/if}}
-		{{#if specialRole}} special-{{specialRole}}{{/if}}
-		{{#if folders}} collapsible{{/if}}
-		{{#if open}} open{{/if}}
-		">
+		class="{{#if unseen}}unread{{/if}}
+			{{#if specialRole}} special-{{specialRole}}{{/if}}
+			{{#if folders}} collapsible{{/if}}
+			{{#if open}} open{{/if}}">
 		{{#if folders}}<button class="collapse"></button>{{/if}}
-		<a class="folder {{#if specialRole}} icon-{{specialRole}}{{/if}}">
+		<a class="folder {{#if specialRole}} icon-{{specialRole}}{{/if}}"
+			tabindex="0">
 			{{name}}
 			{{#if unseen}}
 			<span class="utils">{{unseen}}</span>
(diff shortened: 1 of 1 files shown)


Xenopathic:
The tab index stuff works nicely. To get the pressing enter to work, we will need to listen for the keydown event (which correctly fires on that element, at least in Firefox). I'd suggest listening for spacebar too - they often share the same meaning as far as selecting something is concerned.

jancborchardt:
@Xenopathic can you add that? (I was under the impression that there was also a way to have the markup that it just works.)

Xenopathic:
There is a markup way of doing this... we need to use `<button>` instead of `<a>`. We really should be using `<button>` anyway, these are semantically buttons, not links. Using `<button>` also has implicit tabindex, interestingly enough.

I just tried changing the template to have buttons instead of anchors. The styling was a mess, but tabbing worked and pressing enter triggered the click action.

jancborchardt:
@Xenopathic aaah nice! Cool, I’ll work on changing that and fixing the styling then.

jancborchardt:
Hm, thinking about it more it doesn’t seem to make sense to transform all the elements into buttons. Because currently they already are `a` link elements – that should normally suffice, right? @Volker-E do you have some insight on this?

DeepDiver1975:
@jancborchardt rebase please

Volker-E:
@Xenopathic @jancborchardt I don't have enough time/insight to get into details here right now :\
But I don't see from current understanding a clear a11y advantage of using a `button` element over an `a` in this case? What's your thoughts on that, @Xenopathic ?

Xenopathic:
@Volker-E a11y software looks through the semantic representation of the web page, aka its HTML, to find meaning in the sections. As such, it expects links to other pages to be `<a>` elements, and buttons that do something to be `<button>`s (or `<input>`s, in a form). Using `<button>` also makes everything else fall into place will very little effort - tabbing 'just works' without any tabindex stuff, pressing Enter when the button is selected triggers the onclick action, etc.

biolumin:
I'd be happy to help with this issue. Is there a firm decision on which type of element we'll use? I suggest we use buttons.

If we move forward with changing <a> elements to <button> elements:
I took a look at the code, and it appears the <a> elements are styled from the core's apps.css, even thought mail's mail.css also has an <a> element style, which is currently being ignored. Should we simply add an !important directive and work with the mail.css, or continue to work with the apps.css and find a cohesive solutions?

PR #565, opened 2015-05-01
fix layout of message list on narrower screens and mobile, do not show date for now

jancborchardt:
The date was barely readable anyway because of overlap with the sender. So for now we don’t show it on narrow / mobile screens at all so that at least the sender is readable.

Please review @Xenopathic @zinks- @PoPoutdoor @tomneedham

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -250,6 +250,7 @@
 	position: absolute;
 	right: 0;
 	top: 0;
+	max-width: 40%;
 	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=50)";
 	filter: alpha(opacity=50);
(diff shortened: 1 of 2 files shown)


jancborchardt:
I’m thinking it would rather be better to define a minimum width of the message list, for example `250px` just like for the sidebar. However that collides with the 30%/70% sizing of the content. @wurstchristoph @owncloud/designers @colmoneill any idea?

Xenopathic:
:+1: Looks good, although it starts overlapping before the screen is small enough to cause the sidebar and the dates to disappear - at which point the sidebar disappears giving more space for the message list...

jancborchardt:
Yeah, we need to fix that properly in the future. :)

PR #571, opened 2015-05-02
reposition message indicators, especially star, fix #570

### LINKED ISSUE #570: Confusion between Mail and Contacts app
Because of the usage of the 'contact monograms' in the inbox list, I thought that 'starring' was going to place that contact into my favourite contacts, not add it to my favourite emails.

screenshot from 2015-05-01 23 31 31

I guess this is more a tribute to the platform than anything else, but it's because I'm so used to seeing my contacts listed with their initials this way, I thought for a few seconds that it would interact with the Contacts app.

Just to be clear, I'm definitively not looking to be able to favourite a contact from within Mail, I'm just pointing out (my &) the potential confusion.

Should the single email interaction tools —favourite, delete, reply, reply all, etc— not be moved to the open email panel?

jancborchardt:
Positioning the star left of the title instead of on the user avatar makes it clearer that it will star the message instead of the contact.

Fix https://github.com/owncloud/mail/issues/570, please review @colmoneill @Gomez @wurstchristoph @zinks- :)

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -408,8 +408,8 @@
 	background-size: 16px;
 	display: inline-block;
 	position: absolute;
-	top: 0;
-	left: 31px;
+	left: 25px;
+	top: 37px;
 	z-index: 10;
 }
 /* only show star on hover of row */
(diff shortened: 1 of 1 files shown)


Xenopathic:
It looks odd in that position - it sticks out and looks weird. What about putting it at the far right, above the delete button? Or is usability a concern there?

jancborchardt:
Yes, most definitely a usability concern because delete is there. Can you show a screenshot of how it looks weird? Maybe just fixing the position by a few pixels is better.

Xenopathic:
Well I suspect it looks the way you intend it to look, but right there (underneath the user avatar) just feels wrong. Better than it was before, but still not good.

jancborchardt:
@Xenopathic yeah, I’m not sure either However it seems like the best position – left of the subject of the email, not separate. Can you think of a better position?

But maybe it’s just best to merge it for now cause it’s better than the current situation and then improve if we know something better.

wurstchristoph:
The new position looks definitely better than the old one :+1:

jancborchardt:
Great! As said above I’ll merge that for now then since it’s definitely better than before. We can do further improvements in the future.

colmoneill:
Agreed, this removes the confusion I experienced in #570.

I am expecting this to cause issues though; the attachment icon is informative, starring and deleting are user actions.

How about this: 
screenshot from 2015-05-03 20 28 45

Rather than a verbose time of reception, we shorten it to a predictable amounts of characters, (this will have to be tested for receptions >24hrs something like 30 Apr for the short date format) this saves space makes space for attachment info and favourite option on the top line.

I feel like this could make for a good distinction: top line is informative: Sender, email info, time. Second line is active: subject title and deletion.

This is all the info that will appear in this column, yes? It's a tight space, but we can make it work this way no?

Xenopathic:
@colmoneill That's what I had in mind in my comments above :laughing: Only problem is, when the date disappears due to a small screen, the star will be directly above delete, potentially posing a UX problem?

colmoneill:
Agreed, this could be problematic. I'm wondering if some of these actions could not be moved to the subject bar area in the mail, for mobile sizes.

PR #574, opened 2015-05-03
Expunge mails from the trash

irgendwie:
new pull request for #339

### DIFF
--- a/js/views/message.js
+++ b/js/views/message.js
@@ -2,6 +2,8 @@
 
 var views = views || {};
 
+$('.action.delete').tipsy({gravity:'e', live:true});
+
 views.DetailedMessage = Backbone.Marionette.ItemView.extend({
 	template: "#mail-message-template"
 });
@@ -85,7 +87,9 @@ views.Message = Backbone.Marionette.ItemView.extend({
 		event.stopPropagation();
 		var thisModel = this.model;
 		this.ui.iconDelete.removeClass('icon-delete').addClass('icon-loading');
+		$('.tipsy').remove();
 		this.$el.addClass('transparency').slideUp(function() {
+			$('.tipsy').remove();
 			var thisModelCollection = thisModel.collection;
 			var index = thisModelCollection.indexOf(thisModel);
 			var nextMessage = thisModelCollection.at(index+1);
--- a/lib/account.php
+++ b/lib/account.php
@@ -308,11 +308,19 @@ class Account {
 		$hordeSourceMailBox = new Horde_Imap_Client_Mailbox($sourceFolderId);
 		$hordeTrashMailBox = new Horde_Imap_Client_Mailbox($trashId);
 
-		$result = $this->getImapConnection()->copy($hordeSourceMailBox, $hordeTrashMailBox,
-			array('create' => $createTrash, 'move' => true, 'ids' => $hordeMessageIds));
+		if ($sourceFolderId === $trashId) {
+			$this->getImapConnection()->expunge($hordeSourceMailBox,
+				array('ids' => $hordeMessageIds, 'delete' => true));
 
-		\OC::$server->getLogger()->info("Message moved to trash: {message} from mailbox {mailbox}",
-			array('message' => $messageId, 'mailbox' => $sourceFolderId));
+			\OC::$server->getLogger()->info("Message expunged: {message} from mailbox {mailbox}",
+				array('message' => $messageId, 'mailbox' => $sourceFolderId));
+		} else {
+			$this->getImapConnection()->copy($hordeSourceMailBox, $hordeTrashMailBox,
+				array('create' => $createTrash, 'move' => true, 'ids' => $hordeMessageIds));
+
+			\OC::$server->getLogger()->info("Message moved to trash: {message} from mailbox {mailbox}",
+				array('message' => $messageId, 'mailbox' => $sourceFolderId));
+		}
 	}
 
 	/**
--- a/templates/index.php
+++ b/templates/index.php
@@ -75,7 +75,7 @@ script('mail', 'jquery-visibility');
 						{{relativeModifiedDate dateInt}}
 					</span>
 			</div>
-			<div class="icon-delete action delete"></div>
+			<div class="icon-delete action delete" title="{{delete}}"></div>
 		</div>
 	</div>
 </script>
--- a/lib/controller/messagescontroller.php
+++ b/lib/controller/messagescontroller.php
@@ -93,6 +93,10 @@ class MessagesController extends Controller
 
 		$ci = $this->contactsIntegration;
 		$json = array_map(function($j) use($ci, $mailBox) {
+			if ($mailBox->getSpecialRole() === 'trash') {
+				$j['delete'] = (string)$this->l10n->t('Delete permanently');
+			}
+
 			if ($mailBox->getSpecialRole() === 'sent') {
 				$j['fromEmail'] = $j['toEmail'];
 				$j['from'] = $j['to'];
@@ -100,6 +104,7 @@ class MessagesController extends Controller
 					$j['from'] .= ' ' . $this->l10n->t('& others');
 				}
 			}
+
 			$j['senderImage'] = $ci->getPhoto($j['fromEmail']);
 			return $j;
 		}, $json);


lib/account.php
```diff
@@ -292,11 +292,19 @@ public function deleteMessage($sourceFolderId, $messageId) {
 		$hordeSourceMailBox = new Horde_Imap_Client_Mailbox($sourceFolderId);
 		$hordeTrashMailBox = new Horde_Imap_Client_Mailbox($trashId);
 
-		$result = $this->getImapConnection()->copy($hordeSourceMailBox, $hordeTrashMailBox,
-			array('create' => $createTrash, 'move' => true, 'ids' => $hordeMessageIds));
+		if ($hordeSourceMailBox == $hordeTrashMailBox) {
```
Xenopathic: This check could be potentially dangerous, if the values inside the objects get out of sync. Perhaps check if the IDs are equal, since they are strings? Also, you can then use the safer `===` operator.

lib/account.php
```diff
@@ -292,11 +292,19 @@ public function deleteMessage($sourceFolderId, $messageId) {
...
+			$result = $this->getImapConnection()->expunge($hordeSourceMailBox,
+				array('ids' => $hordeMessageIds, 'delete' => true));
 
-		\OC::$server->getLogger()->info("Message moved to trash: {message} from mailbox {mailbox}",
-			array('message' => $messageId, 'mailbox' => $sourceFolderId));
+			\OC::$server->getLogger()->info("Message expunged: {message}",
+				array('message' => $messageId));
```
Xenopathic: Log the mailbox here too, like below (aids debugging)

irgendwie:
@Xenopathic changed it.

>I would say it’s ok for the first pass to not have an undo or warning there. Maybe what suffices is to use a tipsy tooltip which says »Delete permanently« and pops up when you hover over the delete icon.
>
>You would just add the title via the `title` attribute, and then invoke tipsy in the JS via: `$('class of the delete function').tipsy({gravity:'e', live:true});`

Couldn't find any reference implementation in the mail app and failed to implement this feature @jancborchardt. Some help there? :-)
Also, wouldn't I need the information that I'm in the trash folder, like with the icon approach?

jancborchardt:
Right, we need the information of being in the trash folder in any case.

@DeepDiver1975 how can we find that out in the Javascript? Please advise. :)

DeepDiver1975:
> Right, we need the information of being in the trash folder in any case.
> 
> @DeepDiver1975 how can we find that out in the Javascript? Please advise. :)

let me see ...

DeepDiver1975:
bildschirmfoto von 2015-05-03 21 33 28

DeepDiver1975:
```js
Mail.State.folderView.collection.get(Mail.State.currentAccountId).get('folders').get(Mail.State.currentFolderId).get('specialRole')

**Comment by irgendwie** [comment]
Thanks @DeepDiver1975! Could you review 99ef579869ac27d321c66be28009a42b1977e609, if this will work out?

**Comment by irgendwie** [comment]
@DeepDiver1975 @Xenopathic recheck pls

**Comment by DeepDiver1975** [comment]
Looks really good - and works :+1:

**Comment by DeepDiver1975** [comment]
@jancborchardt these bloody tipsies don't go away - any idea? THX

File: lib/account.php [hunk 3]
```diff
@@ -292,11 +292,19 @@ public function deleteMessage($sourceFolderId, $messageId) {
 		$hordeSourceMailBox = new Horde_Imap_Client_Mailbox($sourceFolderId);
 		$hordeTrashMailBox = new Horde_Imap_Client_Mailbox($trashId);
 
-		$result = $this->getImapConnection()->copy($hordeSourceMailBox, $hordeTrashMailBox,
-			array('create' => $createTrash, 'move' => true, 'ids' => $hordeMessageIds));
+		if ($sourceFolderId === $trashId) {
+			$result = $this->getImapConnection()->expunge($hordeSourceMailBox,
```
DeepDiver1975: remove $result - not needed - THX

lib/account.php
```diff
@@ -292,11 +292,19 @@ public function deleteMessage($sourceFolderId, $messageId) {
...
+				array('ids' => $hordeMessageIds, 'delete' => true));
 
-		\OC::$server->getLogger()->info("Message moved to trash: {message} from mailbox {mailbox}",
-			array('message' => $messageId, 'mailbox' => $sourceFolderId));
+			\OC::$server->getLogger()->info("Message expunged: {message} from mailbox {mailbox}",
+				array('message' => $messageId, 'mailbox' => $sourceFolderId));
+		} else {
+			$result = $this->getImapConnection()->copy($hordeSourceMailBox, $hordeTrashMailBox,
```
DeepDiver1975: remove $result - not needed - THX

DeepDiver1975:
@irgendwie can I ask you to squash your commits? THX

templates/index.php
```diff
@@ -75,7 +75,7 @@ class="
 						{{relativeModifiedDate dateInt}}
 					</span>
 			</div>
-			<div class="icon-delete action delete"></div>
+			<a class="icon-delete delete action" original-title="{{delete}}"></a>
```
Xenopathic: Don't use `original-title` here, just `title` (`original-title` is created by Tipsy, and if we switch to Bootstrap at some point it might change)

jancborchardt:
Where is the tipsy called? Just the one from core js.js on the delete class? Maybe it's because original-title needs to be changed to title.

DeepDiver1975:
> Where is the tipsy called? Just the one from core js.js on the delete class? Maybe it's because original-title needs to be changed to title.

answered by @Xenopathic https://github.com/owncloud/mail/pull/574#discussion_r29560826

irgendwie:
Ok... now... changed accordingly and squashed the commits!

Xenopathic:
Hmm... I can see that the correct code is being fired, but I'm getting that old error again... `Could not open mailbox \"\"."`. So I can't actually test if the message gets deleted (but it would do if Horde didn't mess up).

However, a more pressing issue is that we have no Tipsy's at all! The title correctly shows a tooltip when hovering over the Delete icon, but it is a standard browser tooltip. You might want to dig around core.js and see if there's a quick class you can apply to get Tipsy working again (it's a mess, don't question it :see_no_evil: )

irgendwie:
this should do it @Xenopathic

DeepDiver1975:
the tipsy will permanently stay visible a soon as a message is deleted from trash ... looking int this

DeepDiver1975:
lets add this to the deleteMessage function

https://github.com/owncloud/mail/blob/master/js/views/message.js#L87
```js
$('.tipsy').remove()

**Comment by irgendwie** [comment]
@DeepDiver1975 done, thanks!

**Comment by jancborchardt** [comment]
@irgendwie nice! Can you now squash the commits into one for more clean history? :)

**Comment by irgendwie** [comment]
@jancborchardt done, can be merged!

**Comment by jancborchardt** [comment]
@irgendwie there’s still a slight issue with the tipsy: When you click the delete icon and _not_ move your mouse (or just a little), the next message will move into view. That will result in the mouse hovering over the delete action of the next view, calling the tipsy and also showing it permanently (til the next delete). Maybe the end of the delete animation also needs the tipsy kill function.

File: js/views/message.js [hunk 6]
```diff
@@ -85,6 +87,7 @@ views.Message = Backbone.Marionette.ItemView.extend({
 		event.stopPropagation();
 		var thisModel = this.model;
 		this.ui.iconDelete.removeClass('icon-delete').addClass('icon-loading');
+		$('.tipsy').remove()
```
DeepDiver1975: we need to move this inside the sideup function below - line 98

jancborchardt:
And another thing: We should _not_ have any tipsy for Delete in other folders. Because that will make the important »Delete permanently« notice in the trash less important.

js/views/message.js
DeepDiver1975: and please add the missing ; :speak_no_evil: THX

js/views/message.js
```diff
@@ -85,7 +87,9 @@ views.Message = Backbone.Marionette.ItemView.extend({
 		event.stopPropagation();
 		var thisModel = this.model;
 		this.ui.iconDelete.removeClass('icon-delete').addClass('icon-loading');
+		$('.tipsy').remove();
```
jancborchardt: No need for this tipsy.remove when it will be called 2 lines after anyway. ;)

jancborchardt:
:+1: just waiting for Travis

zinks-:
:+1: nice @irgendwie !

wurstchristoph:
I have the same issues as @jancborchardt  (on FF): When I delete a message and don't move my mouse the tipsy shortly flashes (some pixels underneath it should be) and is hidden then although the next message's trash icon is hovered. To get the tipsy again I have to move out of the trash icon and move back in.

jancborchardt:
@wurstchristoph that issue should be fixed since we call tipsy.remove twice. Try refreshing your browser cache (Ctrl+Shift+r)

jancborchardt:
@irgendwie aaand merged! :) Big thank you for the contribution"

wurstchristoph:
@jancborchardt nope, it still doesn't work as expected.

I tested also with Chrome, there I have to move the mouse a little bit to get the tipsy of the next message :/

jancborchardt:
@wurstchristoph that’s strange – I tested it multiple times in both Firefox and Chrome. Can you check again with an incognito browser or different device?

jancborchardt:
@wurstchristoph aaah I get it now! Yes, that also happens for me actually. Can you open a new issue for that?

PR #576, opened 2015-05-03
Skip IMAP tests if credentials are unavailable

Xenopathic:
It's annoying to see a PR marked as failed because the (encrypted) credentials sent to Travis aren't available for forks. Tests are marked as skipped instead of throwing an exception in this case. It's similar to not having the required extensions for memcache tests, as a comparison.

This PR is coming from my fork to test it.

### DIFF
--- a/tests/imap/abstracttest.php
+++ b/tests/imap/abstracttest.php
@@ -19,13 +19,13 @@ abstract class AbstractTest extends \PHPUnit_Framework_TestCase {
 
 	public static function setUpBeforeClass() {
 		if (false === \getenv('EMAIL_USER')) {
-			throw new \RuntimeException(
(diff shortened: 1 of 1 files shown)


tests/imap/abstracttest.php
```diff
@@ -19,13 +19,13 @@
 
 	public static function setUpBeforeClass() {
 		if (false === \getenv('EMAIL_USER')) {
-			throw new \RuntimeException(
+			$this->markTestSkipped(
```
DeepDiver1975: no $this in static context

Xenopathic: Wow, I didn't even notice the static context here...

clementhk:
R-b: me

PR #581, opened 2015-05-04
get mailbox status once for all boxes

DeepDiver1975:


DeepDiver1975:
@jancborchardt

jancborchardt:
:+1: everything still works

lib/account.php
```diff
@@ -123,6 +123,7 @@ public function createMailbox($mailBox) {
 		$conn->createMailbox($mailBox);
 		$this->mailboxes = null;
 
+//		$mailBox = \Horde_Imap_Client_Mailbox::get($mailBox, false)->utf7imap;
```
clementhk: I wonder where this is from, as I thought I've removed all utf7imap related code.

clementhk: Ohh this is a new line, not commenting out old lines..
We should remove this as it makes no sense, I think.

DeepDiver1975: indeed

jancborchardt: Just remove it ;) pull at https://github.com/owncloud/mail/pull/599

PR #587, opened 2015-05-04
add Mailvelope link to settings

jancborchardt:
Add link pointing to Mailvelope so people can use encrypted emails. Please review @DeepDiver1975 @toberndo @uggrock

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -574,7 +574,20 @@
 	margin: 0 auto;
 }
 
-
+/* settings */
+.mailaccount-list .mail-account-name {
+	display: inline-block;
+	width: 90%;
+	overflow: hidden;
+	text-overflow: ellipsis;
+}
+.mailaccount-list .actions {
+	opacity: .5;
+}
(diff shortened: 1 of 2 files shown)


irgendwie:
Functional :+1:, still it's not very pretty :grin:

irgendwie:
How about adding a little lock icon next to "Add mail account", which shows the information in a tooltip or something? Instead of the plain text under the button?

zinks-:
Works, maybe we can pretty it up later? The account list should be clearly separated from it.
 (but :+1:)

DeepDiver1975:
@jancborchardt rebase please

DeepDiver1975:
rebased

DeepDiver1975:
I'm all in to merge this - we can still style it differently later :+1:

jancborchardt:
Yeah, I also think we can style it up later. Just having an icon with tooltip is too little I think. It’s the settings section after all.

PR #592, opened 2015-05-04
fix opacity & shading of several icons

jancborchardt:
Yeah I know @colmoneill maybe CSS preprocessing wouldn’t be such a bad idea. ;)

Please review @wurstchristoph @irgendwie @zinks- @Gomez

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -265,8 +265,8 @@
 	bottom: 0;
 	right: 0;
 	padding: 20px;
-	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=20)";
-	opacity: .2;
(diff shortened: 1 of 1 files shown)


zinks-:
Looks very nice, but this just struck me : why do we keep the 'delete' icon always visible and not the star ?

irgendwie:
Looks good! :+1:

jancborchardt:
@zinks- because the star icon is also an indicator and would otherwise only distinguish itself by color – not good for colorblind people (or also for everyone, because you need to actually look at the icon instead of only glancing). On top of that, delete is more often used than favoriting, and it needs to be usable on touchscreens – hence always shown instead of only on hover. Starring however suffices when only available for the currently active mail.

jancborchardt:
Only CSS –> not covered by tests –> merge. :)

PR #596, opened 2015-05-05
Cleanup deleteMessage()

clementhk:
@DeepDiver1975 

I was wrong in your pr, that it is using guessBestMailBox() already, but here's some cleanup during the read through.
Feel free to reject it as it has nothing to do with fixing, but just some small optimizes.


clementhk:
I'll make a pull request towards your branch instead.

PR #601, opened 2015-05-05
fix account form being autofilled with username and password in Chrome & Safari

### LINKED ISSUE #566: Autofill of the email settings form
These were the values that autofilled when adding an account to the mail app.

screenshot from 2015-05-01 22 35 44

I had to go and delete the autofilled values to find out what the form actually wanted, and it filled in based on my oC account, which are not the same as my external mail settings.

jancborchardt:
fix #566 

Please review @colmoneill @blizzz @Xenopathic @wurstchristoph @irgendwie 

@blizzz thank you for the code! :) As you see you can improve yours by using the standard `.hidden-visually` class from core, as well as adding `tabindex="-1"` so it doesn’t interfere with keyboard navigation.

### DIFF
--- a/templates/index.php
+++ b/templates/index.php
@@ -267,6 +267,14 @@ script('mail', 'jquery-visibility');
 	</div>
 
 	<form id="mail-setup" class="hidden" method="post">
+		<div class="hidden-visually">
+			<!-- Hack for Safari and Chromium/Chrome which ignore autocomplete="off" -->
+			<input type="text" id="fake_user" name="fake_user"
+				autocomplete="off" tabindex="-1">
+			<input type="password" id="fake_password" name="fake_password"
+				autocomplete="off" tabindex="-1">
+		</div>
+
 		<fieldset>
 			<div id="emptycontent">
 				<div class="icon-mail"></div>


irgendwie:
Can **not** reproduce this bug with "Chromium 42.0.2311.135".

jancborchardt:
@irgendwie does this pull request fix the bug for both browsers?

irgendwie:
Dunno, I'm using neither of them two. :see_no_evil:
Feedback from @colmoneill is needed.

jancborchardt:
Also cc @LukasReschke for Safari testing.

Xenopathic:
I'll try to test this (as soon as I'm finished with this physics lesson...)

Xenopathic:
Works great :+1:

Interesting thing I noticed though (@irgendwie this might be why it didn't work for you): the autofill doesn't happen if the email field already has a value, such as when an email address is set in your Personal settings. The IMAP Password box remains blank in this case.

jancborchardt:
@Xenopathic yep, forgot to note that. The issue only appears in the first place if the email field is blank. (That’s how I found https://github.com/owncloud/core/issues/16057 ;)

LukasReschke:
This is a huge hack and only will work until the browsers change their behaviour. A better fix is what we do in core where we use an explicit store credential button for example for the mail settings.

Without something like this this can pretty easily leak your password to a service provider withour you knowing…

jancborchardt:
@LukasReschke I’m not entirely sure I know what you mean. What do we do in core which is comparable to this? And how could your password be leaked here – the hidden boxes are not sent anywhere.

PR #613, opened 2015-05-05
prefill cc field

### LINKED ISSUE #609: Regression: cc field not prefilled on replies
Heyhey @wurstchristoph it seems your commit e97e87480a856c33cbc0e7431fb9eabcf9357393 broke the display/prefill of cc when replying. ;) The field is shown, but not correctly filled in.

cc @DeepDiver1975

wurstchristoph:
fixes #609

### DIFF
--- a/templates/index.php
+++ b/templates/index.php
@@ -156,7 +156,7 @@ script('mail', 'jquery-visibility');
 			class="hidden"
 			{{/unless}}
 			>
-				<input type="text" name="cc" id="cc" value="{{replyCc}}" class="recipient-autocomplete"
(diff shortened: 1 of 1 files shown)


jancborchardt:
Nice, works! :+1:

I only wonder how the commit of yours I mentioned in #609 caused that problem. :D

clementhk:
Hi,

Problem with this pr is that cc from local storage draft will no longer be filled.

clementhk:
IMO you need to also make local storage draft use list in cc then fill in replyCcList.

jancborchardt:
@wurstchristoph can you fix that what @clementhk noted? Thanks! :)

@clementhk good catch!

PR #622, opened 2015-05-06
autosize textarea on send, fix #562

### LINKED ISSUE #562: Automatically reduce size of textarea again after sending
Currently when you wrote a long answer to an email, the textarea is very tall (as it should be). However after successfully sending it should collapse to the default height again.

I tried that by calling `$('.reply-message-body').autosize();` after the value of the textarea is emptied but it doesn’t really work. :\

Any ideas @owncloud/designers?

jancborchardt:
fix #562

Please review @clementhk @colmoneill @Xenopathic @Gomez! :)

### DIFF
--- a/js/send-mail.js
+++ b/js/send-mail.js
@@ -125,7 +125,8 @@ $(function () {
 				OC.Notification.showTemporary(t('mail', 'Message sent!'));
 
 				// close reply
-				$('.reply-message-body').val('');
+				replyMessageBody.val('');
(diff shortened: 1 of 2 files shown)


clementhk:
LGTM

jancborchardt:
@clementhk don’t only look at the code – also git checkout the branch and test it please with your installation. ;) We can’t only do reviews from looking at the code so close to release. ;)

Xenopathic:
Works like a charm :+1: (no doge this time though)

PR #632, opened 2015-05-06
on failed connection activate send/reply button again

### LINKED ISSUE #606: On mail sending error, show more permanent error and enable send button again
When there’s an error reaching the IMAP server, the error vanishes too quickly. I sent a message, switched to another app, then when I switched back the text was still in the box and the send button was deactivated. I needed to check the log to confirm that it was a connection issue.

In that case, the yellow oc notification should show up top and the send button should be activated again. cc @DeepDiver1975 @wurstchristoph @zinks-

jancborchardt:
fix #606. The notification vanishing when in the background is a core issue, see https://github.com/owncloud/core/issues/16121

Please review @DeepDiver1975 @wurstchristoph @Xenopathic @irgendwie 

To test:
1. Load mail app
2. Compose reply or new message
3. Deactivate your wifi
4. Press »send« or »reply«
5. See the error and see the buttons being active again

### DIFF
--- a/js/send-mail.js
+++ b/js/send-mail.js
@@ -108,6 +108,7 @@ $(function () {
 			complete:function () {
 				replyMessageBody.removeClass('icon-loading');
 				replyMessageBody.prop('disabled', false);
+				replyMessageSend.prop('disabled', false);
 				replyMessageSend.val(t('mail', 'Reply'));
 				$('.reply-message-fields #to').prop('disabled', false);
 				$('.reply-message-fields #cc').prop('disabled', false);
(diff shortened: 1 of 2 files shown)


Xenopathic:
:+1: tested, works, code looks good

irgendwie:
:+1: tested, works as expected!

clementhk:
LGTM

jancborchardt:
Merging and cancelling Travis because it was reviewed by 3. Travis takes too long and clogs up our pipeline.

PR #638, opened 2015-05-06
load multi level subfolders

zinks-:
Now subfolders of subfolders can be loaded
Sent
 > Foo
 > Foo/Bar
 > Foo/Bar/barfoo

The window title says "undefined", but still better than not loading the messages!

### DIFF
--- a/js/views/folder.js
+++ b/js/views/folder.js
@@ -87,6 +87,11 @@ views.Folders = Backbone.Marionette.CollectionView.extend({
 				k += '/';
 			}
 			k += p;
+
+			if (!_.isUndefined(activeFolder)) {
+				return;
+			}
+
 			var folders = activeFolder.folders || activeFolder.get('folders');
 			activeFolder = folders.filter(function(f) {
 				return f.id === btoa(k);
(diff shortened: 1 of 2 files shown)


zinks-:
@jancborchardt @DeepDiver1975 @Xenopathic 
Should be able to load any depth of subfolders with all directory separators, please test

js/views/folder.js
```diff
@@ -83,10 +82,15 @@ views.Folders = Backbone.Marionette.CollectionView.extend({
 		var activeFolder = activeAccount;
```
Xenopathic: You'll probably want to use the mail delimiter on the line above this one too (`folderId.split()`)

zinks-:
@Xenopathic thanks for noticing, fixed

zinks-:
closing this for now as there are so many things wrong with what I'm doing here !

Xenopathic:
@zinks- Are there? It seemed quite sensible to me

clementhk:
@zinks- there are so many things wrong in the original proposal as well :) I guess we'll just have to live with it for now.

PR #640, opened 2015-05-07
coding style improvements

wurstchristoph:
- short array syntax
- double emtpy lines
- opening curly braces
- unused use statements
- space indents

### DIFF
--- a/lib/controller/folderscontroller.php
+++ b/lib/controller/folderscontroller.php
@@ -72,7 +72,7 @@ class FoldersController extends Controller {
 			foreach($folders as &$parent) {
 				if($parent['id'] === $parentId) {
 					if (!isset($parent['folders'])) {
-						$parent['folders'] = array();
+						$parent['folders'] = [];
 					}
 					$parent['folders'][] = $folder;
 					break;
@@ -137,7 +137,7 @@ class FoldersController extends Controller {
 
 			$newFolderId = $mailbox;
 			return new JSONResponse(
-				array('data' => array('id' => $newFolderId)),
+				['data' => ['id' => $newFolderId]],
 				Http::STATUS_CREATED);
 		} catch (\Horde_Imap_Client_Exception $e) {
 			$response = new JSONResponse();
--- a/lib/controller/messagescontroller.php
+++ b/lib/controller/messagescontroller.php
@@ -22,8 +22,8 @@ use OCP\AppFramework\Http\JSONResponse;
 use OCP\AppFramework\Http\TemplateResponse;
 use OCP\IL10N;
 
-class MessagesController extends Controller
-{
+class MessagesController extends Controller {
+
 	/**
 	 * @var \OCA\Mail\Db\MailAccountMapper
 	 */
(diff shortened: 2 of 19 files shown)


wurstchristoph:
we should also get rid of ```_someMethod()``` naming conventions like in https://github.com/owncloud/mail/blob/master/lib/cache/cache.php#L97

what do you think @DeepDiver1975? seems like you introduced those ;-)

lib/controller/messagescontroller.php
```diff
@@ -159,7 +159,7 @@ public function getHtmlBody($messageId) {
 
 			return new HtmlResponse($html);
 		} catch(\Exception $ex) {
-			return new TemplateResponse($this->appName, 'error', array('message' => $ex->getMessage()), 'none');
+			return new TemplateResponse($this->appName,'error', ['message' => $ex->getMessage()], 'none');
```
DeepDiver1975: keep the space please

lib/service/html.php
```diff
@@ -96,7 +95,8 @@ public function __construct() {
 	 * @return string
 	 */
 	public function convertLinks($data) {
-		$regex = "/(ht|f)tp(s?)\:\/\/(([a-zA-Z0-9\-\._]+(\.[a-zA-Z0-9\-\._]+)+)|localhost)(\/?)([a-zA-Z0-9\-\.\?\,\'\/\\\+&amp;%\$#!_]*)?([\d\w\.\/\%\+\-\=\&amp;\?\:\\\&quot;\'\,\|\~\;]*)/";
```
DeepDiver1975: I'd prefer to keep this string in one line - on concat of regex a char is easily lost and your will not recognize - THX

DeepDiver1975:
> we should also get rid of _someMethod() naming conventions like in https://github.com/owncloud/mail/blob/master/lib/cache/cache.php#L97
> 
> what do you think @DeepDiver1975? seems like you introduced those ;-)

that class is a pure copy of a Horde class - I'd prefer to keep it unchanged

DeepDiver1975:
https://github.com/horde/horde/blob/master/framework/Imap_Client/lib/Horde/Imap/Client/Cache/Backend.php

wurstchristoph:
@DeepDiver1975 thanks for your feedback!
Reverted cache.php modifications and fixed the other two things :-)

jancborchardt:
@DeepDiver1975 @zinks- @irgendwie @clementhk  can you review this?

lib/controller/folderscontroller.php
```diff
@@ -137,7 +137,7 @@ public function create() {
 
 			$newFolderId = $mailbox;
 			return new JSONResponse(
-				array('data' => array('id' => $newFolderId)),
+				array('data' => ['id' => $newFolderId]),
```
zinks-: a stray array

wurstchristoph: @zinks- thanks, removed it ;-)

zinks-:
Good calls, although I prefer the array() syntax but that's a matter of taste :+1:

DeepDiver1975:
:+1:

Issue #642, opened 2015-05-08
min OC version


aimlessadam:
i have been watching this project with great anticipation for the past few months, waiting for the 0.1 version to drop.  checking in today, it seemed this would be the day so i left the project root page open in a browser and was hitting refresh periodically... then, during the last refresh, i saw the Readme.md change from saying "An email app for ownCloud (minimum version 7)..." to "An email app for ownCloud (minimum version 8)..." and, sigh, there went my excitement. 

i definitely understand that the world is moving forward, and supporting older versions is costly, but i'm still running a couple apps that are not ready for OC 8. and since it seems like the min version got bumped on release day, i'm posting this to ask if v8 is a hard requirement?

thank you in any case for working on this, OwnCloud sooo needs a native mail client capable of working with files.  This will be quickly the #1 app i predict.

DeepDiver1975:
@aimlessadam Thanks a lot for your feedback! I'm glad to hear about your excitement!

Let me see how much work it will be to make if OC7 compliant!

aimlessadam:
thank you very much!

jancborchardt:
Yeah, maybe it makes sense to see about ownCloud 7 compatibility. Thank you for the feedback @aimlessadam – you are very welcome in our IRC channel #owncloud-mail! :)

zinks-:
an argument against it is that we're only 0.1, and when we hit 1.0 oc7 probably won't be suported anymore
but then.. it all depends on the amount of work to make it compatible for now

clementhk:
Hi,

If we do this, I think we should only support the installations that uses PHP >= 5.4, since there are a lot of changes that requires >= 5.4, but OC 7 only requires PHP 5.3.8.
How shall we deal with this? Is it allow to have a different dependency in Apps?

DeepDiver1975:
We can specify min php version in info.xml - that's not an issue. 

We will see how big the effort is to make the app oc7 compliant.

jancborchardt:
Fixed by https://github.com/owncloud/mail/pull/649, I also updated the readme. Thank you for noting this @aimlessadam! :) You’re very welcome in our IRC channel #owncloud-mail

aimlessadam:
wow i was travelling yesterday, and missed a lot.  you guys were very busy.  thank you.

i have to work on coaxing a php upgrade into my centos-6 today first, but i hope to be trying v0.1 later today or tomorrow.  

thank you again!

jancborchardt:
[...]

PR #649, opened 2015-05-10
Make mail ownCloud7 compliant

DeepDiver1975:
refs #642 

@jancborchardt @zinks- @rullzer @MorrisJobke please review and test - THX

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,8 @@ matrix:
     env: DB=pgsql
   - php: 5.4
     env: "DB=mysql CORE_BRANCH=stable8"
+  - php: 5.4
+    env: "DB=mysql CORE_BRANCH=stable7"
 
   allow_failures:
   - php: hhvm
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -11,7 +11,8 @@
 		<admin>https://github.com/owncloud/mail#readme</admin>
 	</documentation>
 	<dependencies>
-		<owncloud min-version="8.0" max-version="8.1" />
+		<owncloud min-version="7.0" max-version="8.1" />
+		<php min-version="5.4" max-version="5.6" />
 	</dependencies>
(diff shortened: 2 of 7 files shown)


jancborchardt:
Also cc @clementhk @irgendwie @wurstchristoph check this out ^

jancborchardt:
Oh and @aimlessadam please review this as well as it fixes #642 :)

jancborchardt:
@DeepDiver1975 we can also just put the CSS stuff on the bottom of the normal CSS file with a note to be removed if we would drop ownCloud 7 compatibility. That seems a bit simpler than conditionally loading another CSS file. What do you think? I can do it right away.

jancborchardt:
Works fine still on my ownCloud 8 dev setup. But on my (slightly older) ownCloud 7 I get these errors.

First when installing this notice:
```
Invalid OCS content returned for app mail
```
Then the app does not load:
```
Call to undefined method OC\\Server::getCrypto()
at /apps\/mail\/appinfo\/application.php line 121
```

Any idea? Is that because the ownCloud 7 is _too_ old? cc @LukasReschke because crypto.

DeepDiver1975:
Hmm ... might be that we require a certain 7.0.x   .... let me check later

jancborchardt:
The version I have is 7.0.2 (pulled via git).

jancborchardt:
Will also test with a fresh local setup.

jancborchardt:
@DeepDiver1975 works very nicely with latest ownCloud 7! :+1: :) We can’t do anything about search not working, right?

Another review please @aimlessadam @clementhk @zinks- @irgendwie @MorrisJobke! Last one for 0.1 :rocket:

irgendwie:
@jancborchardt Used latest stable7 (from git) and the mail app seems to work as expected - tried some basic stuff! :+1: very nice @DeepDiver1975

Issue #648, opened 2015-05-10
Ability to include original message / inline reply


jancborchardt:
One thing we still do differently by default from other mail apps is that we simply remove the previous reply. That’s one of the reasons I sometimes still fire up another mail app, just to be able to inline reply or to not have the message thread lost for people cc’d after.

We could solve that by having a setting to »include original message«. Either for every message next to the reply button, or globally in the settings.

What it would simply do is copy the previous message, put `>` in front of it and paste it below what you just wrote (or if you didn’t write yet, have 3 empty lines above it). That would also automatically give us inline replies since then you can just go in manually and inline reply.

jancborchardt:
This is currently the _only_ reason why I still sometimes need to start up Geary to reply to an email. ;) @DeepDiver1975 @wurstchristoph @irgendwie @zinks- anyone of you wants to work on this?

irgendwie:
I will take a look the next days. :+1:

jancborchardt:
@irgendwie great stuff! :) (Will mark this as »Developing« because I can’t yet assign you.)

jancborchardt:
Fixed by Alex @irgendwie via #679!

PR #679, opened 2015-05-14
Inline reply functionality

irgendwie:
It's work in progress, but should already work for text messages.
Edit: PR for #648 @jancborchardt

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -321,7 +321,7 @@ var Mail = {
 				Mail.State.folderView.changeUnseen);
 
 			// request permissions
-			if(typeof Notification !== 'undefined') {
+			if (typeof Notification !== 'undefined') {
 				Notification.requestPermission();
 			}
 
@@ -631,6 +631,18 @@ var Mail = {
 					data.replyCc = draft.cc;
 					data.replyBcc = draft.bcc;
 					data.replyBody = draft.body;
+				} else {
+					// Add body content to inline reply (text mails)
+					if (!data.hasHtmlBody) {
+						var date = new Date(data.dateIso);
+						var minutes = date.getMinutes();
+
+						data.replyBody = '\n\n\n\n' +
+							data.from + ' – ' +
+							$.datepicker.formatDate('D, d. MM yy ', date) +
+							date.getHours() + ':' + (minutes < 10 ? '0' : '') + minutes + '\n> ' +
+							jQuery(data.body).text().replace(/\n/g, '\n> ');
+					}
 				}
 
 				// Render the message body
@@ -675,10 +687,32 @@ var Mail = {
 					});
 					// Remove spinner when loading finished
 					$('iframe').parent().removeClass('icon-loading');
+					// Add body content to inline reply (html mails)
+					// TODO better html handling, should be fine for the moment!
+					var text = jQuery($(this).contents().find('body').html().replace(/<br>/g, '\n')).text();
+					if (!draft) {
+						var date = new Date(data.dateIso);
+						var minutes = date.getMinutes();
+
+						$('.reply-message-body').first().text(
+							'\n\n\n' +
+							data.from + ' – ' +
+							$.datepicker.formatDate('D, d. MM yy ', date) +
+							date.getHours() + ':' + (minutes < 10 ? '0' : '') + minutes + '\n> ' +
+							text.replace(/\n/g, '\n> ')
+						);
+					}
 
 				});
 
 				$('textarea').autosize({append: '"\n\n"'});
+				// dirty workaround to set reply message body to the default size
+				$('.reply-message-body').css('height', '');
+
+				// Expand reply message body on click
+				$('.reply-message-body').click(function() {
+					$('.reply-message-body').trigger('autosize.resize');
+				});
 			};
 
 			var loadDraftSuccess = function(data) {
@@ -748,7 +782,7 @@ var Mail = {
 		},
 
 		toggleSendButton:function () {
-			if(($('#to').val() !== '') && (($('#subject').val() !== '') || ($('#new-message-body').val() !== ''))) {
+			if (($('#to').val() !== '') && (($('#subject').val() !== '') || ($('#new-message-body').val() !== ''))) {
 				$('#new-message-send').removeAttr('disabled');
 			} else {
 				$('#new-message-send').attr('disabled', true);
@@ -756,7 +790,7 @@ var Mail = {
 		},
 
 		toggleReplyButton:function () {
-			if(($('.reply-message-fields #to').val() !== '') && ($('.reply-message-body').val() !== '')) {
+			if (($('.reply-message-fields #to').val() !== '') && ($('.reply-message-body').val() !== '')) {
 				$('.reply-message-send').removeAttr('disabled');
 			} else {
 				$('.reply-message-send').attr('disabled', true);
@@ -766,7 +800,7 @@ var Mail = {
 		toggleManualSetup:function() {
 			$('#mail-setup-manual').slideToggle();
 			$('#mail-imap-host').focus();
-			if($('#mail-address').parent().prop('class') === 'groupmiddle') {
+			if ($('#mail-address').parent().prop('class') === 'groupmiddle') {
 				$('#mail-password').slideToggle(function() {
 					$('#mail-address').parent()
 						.removeClass('groupmiddle').addClass('groupbottom');
--- a/lib/message.php
+++ b/lib/message.php
@@ -394,6 +394,7 @@ class Message {
 		$data['size'] = Util::humanFileSize($this->getSize());
 		$data['flags'] = $this->getFlags();
 		$data['dateInt'] = $this->getSentDate()->getTimestamp();
+		$data['dateIso'] = $this->getSentDate()->format('c');
 		$data['ccList'] = $this->getCCList();
 		return $data;
 	}


js/mail.js
```diff
@@ -627,6 +627,17 @@ var Mail = {
...
+				} else {
+					if(data.hasHtmlBody) {
+						/*// Doesn't work here, because html of mail is rendered later
+						// Needs info! also
+						// TODO language support
+						jQuery.get(data.htmlBodyUrl, function(text) {
+							data.replyBody = '> ' + jQuery(text).text().replace(/\n/g, '\n> ');
+						});*/
+					} else {
+						data.replyBody = '\n\nOn ' + data.date + ' ' + data.from + ' wrote:\n> ' + jQuery(data.body).text().replace(/\n/g, '\n> ');
```
jancborchardt: We need to make this translatable. But I’m actually thinking if it’s better in a simpler format. Instead of:
* On 17 mai 2015 14:40:18 UTC+2 Jan-Christoph Borchardt wrote:
Better
* Jan-Christoph Borchardt (17 May 2015, 14:40)

Like a chat format.

jancborchardt:
Nice! :) Just the few things:

* there should be two (or even three) blank lines above the quote instead of just one
* check my comment about the different format for the quote header
* yeah, we have to make it work for HTML mails. Otherwise it’s super-inconsistent for which mails it works. Especially as we format most simple HTML mails to look like text mails.

Great work @irgendwie! :) Let’s continue with it on Wednesday.

jancborchardt:
Btw @irgendwie one thing to improve would be more descriptive branch names. :) For example, `inline-reply` instead of `fix-648`.

irgendwie:
@jancborchardt I'll have that in mind for the future ;-)

DeepDiver1975:
> Btw @irgendwie one thing to improve would be more descriptive branch names. :) For example, inline-reply instead of fix-648.

fix-648 is highly descriptive to me

js/mail.js
DeepDiver1975: gmail format:

````
rftgzhujioko

2015-05-10 14:23 GMT+02:00 admin <deepdiver1975@gmail.com>:
dfsfdsfasdasd

Comment (DeepDiver1975) [inline js/mail.js]: In terms of international usage we shall keep the timezone?

**Comment by DeepDiver1975** [comment]
> yeah, we have to make it work for HTML mails. Otherwise it’s super-inconsistent for which mails it works. Especially as we format most simple HTML mails to look like text mails.

we need to dig into the html purifyer to see what we can do - maybe @LukasReschke can help out tomorrow as we meet :rocket:

**Comment by jancborchardt** [comment]
> fix-648 is highly descriptive to me

You confuse »descriptive« with »accurate«. ;) The issue can be mentioned in the commit message. When you do a git branch to show all branches and only see a list of fix-* branches, you have no clue what they are about.

**Comment by DeepDiver1975** [comment]
> you have no clue what they are about.

to be honest - it's quiet simple - they fix that issue .... :wink:

**Comment by DeepDiver1975** [comment]
@irgendwie there are some js issue - see travis

````
> jshint . && ./node_modules/.bin/jsxhint .

js/mail.js: line 633, col 29, Bad line breaking before '+'.

js/mail.js: line 634, col 29, Bad line breaking before '+'.

js/mail.js: line 635, col 29, Bad line breaking before '+'.

js/mail.js: line 683, col 29, Bad line breaking before '+'.

js/mail.js: line 684, col 29, Bad line breaking before '+'.

js/mail.js: line 685, col 29, Bad line breaking before '+'.

js/mail.js: line 634, col 31, 'moment' is not defined.

js/mail.js: line 684, col 31, 'moment' is not defined.

**Comment by irgendwie** [comment]
@jancborchardt Review please! HTML handling should be much better now!

**Comment by jancborchardt** [comment]
This is pretty much awesome now!

Just one little nitpick with the date format (which you will be able to handle in no time now you are a master of moment.js ;):
Leading zeroes in the day of the month and in the hour should be removed. That is »05. May« should be »5. May«. And »04:20« should be »4:20«.

Otherwise :+1: :) (I also checked off the done items in my comment above.)

File: js/mail.js [hunk 2]
```diff
@@ -671,6 +678,14 @@ var Mail = {
 					});
 					// Remove spinner when loading finished
 					$('iframe').parent().removeClass('icon-loading');
+					// Add body content to inline reply
+					var text = $(this).contents().find('body')[0].innerText;
+					document.getElementsByClassName('reply-message-body')[0].innerText = (
+						'\n\n\n' +
+						data.from + ' – ' +
+						moment(data.dateIso).format('ddd, DD. MMM YYYY HH:mm') + '\n> ' +
```
jancborchardt: I suppose the comment I made about the date format would be solved here by using `ddd, D. MMM YYYY H:mm` instead?

irgendwie:
:imp: those... hours... wasted :D
fixed!

jancborchardt:
:+1: I for one welcome this PR as our new overlord.

js/mail.js
```diff
@@ -631,6 +631,14 @@ var Mail = {
 					data.replyCc = draft.cc;
 					data.replyBcc = draft.bcc;
 					data.replyBody = draft.body;
+				} else {
+					// Add body content to inline reply (text mails)
+					if(!data.hasHtmlBody) {
```
jancborchardt: Yep. I also thought before that `if(` was better style but the JS guidelines from both Mozilla and Google say `if (`. :)

jancborchardt:
@DeepDiver1975 @irgendwie I assume the JS issues mentioned above by @DeepDiver1975 are also fixed? https://github.com/owncloud/mail/pull/679#issuecomment-104182353

js/mail.js
```diff
@@ -675,6 +683,16 @@ var Mail = {
 					});
 					// Remove spinner when loading finished
 					$('iframe').parent().removeClass('icon-loading');
+					// Add body content to inline reply (html mails)
+					var text = $(this).contents().find('body')[0].innerText;
```
wurstchristoph: I get an error in FF saying ``TypeError: text is undefined``

I fixed it by using ``var text = $(this).contents().find('body').first().text();``

js/mail.js
```diff
@@ -675,6 +683,16 @@ var Mail = {
...
+					if(!draft) {
+						document.getElementsByClassName('reply-message-body')[0].innerText = (
```
wurstchristoph: This can be replaced by jQuery like ``$('.reply-message-body').first().text(...);`` I guess

jancborchardt:
@wurstchristoph do you have anything against merging this and doing fixes in follow-up PRs? As far as I tested nothing is broken by this PR – we can polish incrementally and shouldn’t let this linger for too long. :)

irgendwie:
`js/mail.js: line 639, col 29, 'moment' is not defined.`
`js/mail.js: line 692, col 29, 'moment' is not defined.`

@DeepDiver1975  any idea on that?

wurstchristoph:
inline replies work very well unless the original message is HTML:


As commented above, using ``var text = $(this).contents().find('body').text();`` fixes the issue for me :confused:

I tested with FF with a github mail and a HTML mail composed with evolution

@jancborchardt I would say this needs to be fixed before it is merged

jancborchardt:
Yeah, we need to fix Travis and the JS there.

@irgendwie does @wurstchristoph’s fix work for you? @biolumin in case you’re up for helping with the JS.

irgendwie:
@wurstchristoph added your suggestions, would you mind reviewing it? :-)
cc @jancborchardt and @DeepDiver1975 (for the `'moment' is not defined`)

wurstchristoph:
@irgendwie the JS error is fixed :-)

However, I found another bug: the textarea element is not resized automatically, I need to click into it and press some key (e.g. space), then the element is resized to show the full content:
# Before

# After pressing a key inside the textarea


It seems that this happens if the content of the textarea is changed after the textarea was created... Maybe jQuery autosize has to be triggered after loading the reply content into the textarea. Maybe https://github.com/owncloud/mail/blob/b7977ae2dbd51c7cda2ea32280538d6e96e08794/js/views/sendmail.js#L149 helps

irgendwie:
@wurstchristoph this is intended - the reply message should be default size at the beginning, so that it is clear, what its purpose is. If the reply message body is extended from the beginning, maybe it isn't clear what it is used for unless you completely scroll down until you see the reply button. It's more or less a feature request of @jancborchardt and I think this sounds reasonable.

I already implemented a resize on click event, so that if you click in the reply message body, the text area is expended. Still, drafts and text messages are completely expanded from the beginning - this needs to be "fixed".

wurstchristoph:
:+1: looks good to me

irgendwie:
@jancborchardt I implemented the size/resize as discussed at the meetup!
Additionally, I added a resize on click event, to expand the reply message body, which I think is an intuitive behavior.
So this last change should solve your last checkbox:
> reply box should only expand once you edit the content. Otherwise the reply button is not visible and it might be confusing.

Some code isn't pretty, but functional :see_no_evil:

jancborchardt:
Nice, good stuff, also on the resize! :) Am mobile atm - what do we do with the momentjs stuff for ownCloud 7 compat? Just catch it and use another simple date function? We shouldn't ship it separately I guess.

irgendwie:
Does this work for everyone?
@jancborchardt @wurstchristoph

wurstchristoph:
@irgendwie great that you could remove the moment.js dependency, works very well :+1:

jancborchardt:
Looks very nice, let's finally get this in! :)

PR #689, opened 2015-05-19
show OC notification when draft was saved successfully

### LINKED ISSUE #685: Drafts: quick feedback that message has been saved as draft
We have an awesome drafts feature, but new users don’t know that they can just leave the composer / reply field and it will be saved.

Ideally when navigating away from the new message or the reply, there could be a OC.Notification.showTemporary to say »Message saved as draft«. This feedback will let them know that the feature exists.

(This has come up in a small usability test I did with a person new to the Mail app.)

cc @irgendwie @biolumin :)

wurstchristoph:
fixes #685

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -1,4 +1,4 @@
-/* global Handlebars, Marionette, Notification, relative_modified_date, formatDate, humanFileSize, views */
+/* global Handlebars, Marionette, Notification, relative_modified_date, formatDate, humanFileSize, views, OC */
 
 if (_.isUndefined(OC.Notification.showTemporary)) {
 
@@ -204,6 +204,8 @@ var Mail = {
 		}
 	},
 	UI: {
+		composerVisible: false,
+
 		renderSettings: function () {
 			var source   = $("#mail-settings-template").html();
 			var template = Handlebars.compile(source);
@@ -407,6 +409,8 @@ var Mail = {
 		},
 
 		loadMessages: function (accountId, folderId, noSelect) {
+			Mail.UI.Events.onComposerLeave();
+
 			if (Mail.State.messagesLoading !== null) {
 				Mail.State.messagesLoading.abort();
 			}
@@ -547,6 +551,8 @@ var Mail = {
 		},
 
 		openComposer: function(data) {
+			Mail.UI.composerVisible = true;
+
 			$('#mail_new_message').prop('disabled', true);
 
 			if (Mail.State.composeView === null) {
@@ -560,6 +566,8 @@ var Mail = {
 				Mail.State.composeView.sentCallback = function () {};
 			} else {
 				Mail.State.composeView.data = data;
+				Mail.State.composeView.hasData = false;
+				Mail.State.composeView.hasUnsavedChanges = false;
 			}
 
 			if (data && data.hasHtmlBody) {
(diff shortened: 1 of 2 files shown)


jancborchardt:
Hm, now it basically continuously shows »Draft saved!« on the top. That’s a bit annoying.

I’d say it should only show when you navigate away from the »New message« or »Reply«. Ok?

wurstchristoph:
I agree that the OC notification can be annoying, however, I think we should inform the user that the message was saved. 
If the notification is only shown when a user navigates away, how could he or she notice that the draft was saved already?

jancborchardt:
@wurstchristoph I would say people expect their stuff to be saved anyway. The only thing that’s missing is feedback. But showing this feedback all the time is too much. It’s enough to show only when navigated away and when people might think their stuff is gone.

wurstchristoph:
@jancborchardt alright, I'll have another look at this tomorrow ;-)

jancborchardt:
Great! :)

biolumin:
I don't think a temporary notification at the top of the screen is going to be easily noticed by all people.

We have unused space where the alerts can remain present and clear for users. An example that I believe we should imitate are the alert notifications for 'Saving..' and 'Saved.' as done in GMail.

Here is their approach:
screen shot 2015-05-19 at 7 17 06 pm

Here is the open space we could utilize to accomplish the same thing, and also pave way for future HTML support.
screen shot 2015-05-19 at 7 14 08 pm

jancborchardt:
It’s not about making people force notice that. Using the example of Gmail, they do pretty much what I suggested: When you leave the new message or reply, they show you a notification up top.

Regarding the HTML formatting example from Gmail: Please note that they vastly simplified this in the new Google Inbox interface:
gmail inbox bottom
compose box bottom

> We have unused space

"Unused space" is not necessarily there to fill it up with stuff. ;)

But in any case we’re veering off the path. This pull is about showing a notification when navigating _outside_ of the draft area, so that bottom bar would not help either way and we need a notification.
Let’s discuss the HTML formatting stuff in a new issue you wanted to open. ;)

jancborchardt:
@aidanamavi @colmoneill you also want to review this?

colmoneill:
No too sure what to think of this to be honest, I tried the version with the notification appearing and disappearing, it didn't bother me, but I was looking to test this out, so not a very good candidate for opinion.

The general oC notification is definitively noticed when it appears, but what I guess could happen is if you have an other (permanent) notification up there like I've seen before (ownCloud 8.0.3 is available [...]) it would not be very visible.

I guess the underlying importance is that the draft is actually saved, which is the case, so whether or not the user is notified about the this, they'll be happy to find drafts in their folders.

I am indeed sure that there will be a talk sooner or later about if / how / where to manage text styling abilities, and looking forwards to discussing this with you all.

For now this looks good to me! :)

aidanamavi:
I agree with the opinion of @colmoneill 

Does it resolve #685 ? Yes.
Can we come up with a better approach? Yes. In a new issue.

Bottom line, this code works well and should be merged.
:+1:

irgendwie:
:+1: works and looks good! Still it needs translation, doesn't it?
and btw, I opened a issue for the direct replies #734

PR #692, opened 2015-05-19
translate special 'Favorites' mailbox

### LINKED ISSUE #690: "Favorites" folder name is not translated
the "Favorites" special folder name is not translated :

https://github.com/owncloud/mail/blob/3e54cf1bd5b6ef265f8d111390f1856b2b84ae3b/lib/searchmailbox.php#L26

wurstchristoph:
fixes #690 

the translated string is casted because otherwise it's type is ```object```... but I have no idea why :confused: 

@zinks- @jancborchardt please review :)

### DIFF
--- a/lib/searchmailbox.php
+++ b/lib/searchmailbox.php
@@ -8,6 +8,7 @@
 
 namespace OCA\Mail;
 
+use OC;
 use Horde_Imap_Client;
 use Horde_Imap_Client_Mailbox;
 use Horde_Imap_Client_Search_Query;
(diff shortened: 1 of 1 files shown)


zinks-:
@wurstchristoph Can you instead set the flagged attribute to the SearchMailBox object so that the name is translated at the same place as the others : https://github.com/wurstchristoph/mail/blob/translate-favorites/lib/mailbox.php#L263

That will also enable us to have other SearchMailboxes with different names in the future

thanks!

lib/searchmailbox.php
```diff
@@ -21,12 +22,13 @@ class SearchMailbox extends Mailbox {
...
 		parent::__construct($conn, $mailBox, $attributes, $delimiter);
-		parent::setDisplayName('Favorites');
+		$l = OC::$server->getL10N('mail');
+		parent::setDisplayName((string) $l->t('Favorites'));
 	}
 
-	public function getMessages($from = 0, $count = 2, $filter = '') {
+	public function getMessages($from=0, $count=2, $filter='') {
 		$query = new Horde_Imap_Client_Search_Query();
 		$query->flag('FLAGGED');
```
wurstchristoph: >  @wurstchristoph Can you instead set the flagged attribute to the SearchMailBox object so that the name is translated at the same place as the others : https://github.com/wurstchristoph/mail/blob/translate-favorites/lib/mailbox.php#L263

@zinks- do you mean here? :confused:

zinks-:
hey @wurstchristoph I just went ahead and created #694, it has the benefits of using more of the general mechanisms we already had in place.

wurstchristoph:
@zinks- great! :)

Issue #206, opened 2014-08-20
Automatic redirect for external links in Firefox


jancborchardt:
Automatic redirect doesn’t work for me in Firefox. It works in Chrome though …

@LukasReschke (carried over from https://github.com/owncloud/mail/pull/192)

jancborchardt:
Also, in Firefox for some reason the redirection page shows the noscript warning (as code) up top:
![capture du 2015-04-17 10 08 33](https://cloud.githubusercontent.com/assets/925062/7199142/88847d3a-e4f2-11e4-9362-fa92d9db61dc.png)

Gomez:
The bug does not get any attention @ mozilla. I voted for the bug.

RandolfCarter:
Getting
```
PHP Fatal error:  Call to undefined method OCP\\AppFramework\\Http\\ContentSecurityPolicy::disallowScriptDomain() in /var/www/owncloud/apps/mail/lib/controller/messagescontroller.php on line 176, referer: https://mydomain/owncloud/index.php/apps/mail
```
when trying to run latest master, and folders except inbox aren't refreshed - is it related to this issue? Is it maybe just a setup issue on my part?

LukasReschke:
If you use core master as well as mail master you need to pull both branches. Stable8 + 7 of core should run fine also.

RandolfCarter:
I tried with core 8.1b2 and mail master branch.

LukasReschke:
> I tried with core 8.1b2 and mail master branch.

Yes. That won't work at the moment as the beta 2 does not contain the needed core changes. You need to apply https://github.com/owncloud/core/issues/16463 for that :-( 

Next 8.1 build should work though! :smile:

RandolfCarter:
Ah of course. I saw that one but didn't realize it was a change to the core repo. Must have been quite tired already ;)

PR #696, opened 2015-05-20
Make redirect in Firefox work again and harden CSP policy

LukasReschke:
This makes the redirect in Firefox work again by using CSP instead of a sandboxed iframe. Unit tests are coming later because the message controller has some hard-coded references to other classes with side-effect such as `\OCA\Mail\Account`. Trying to come up with a fix for that as well in another PR.

Requires https://github.com/owncloud/core/pull/16463 for optimal security. Should work without though.

Fixes https://github.com/owncloud/mail/issues/206

cc @DeepDiver1975 @jancborchardt

### DIFF
--- a/lib/controller/messagescontroller.php
+++ b/lib/controller/messagescontroller.php
@@ -13,11 +13,13 @@
 namespace OCA\Mail\Controller;
 
 use Horde_Imap_Client;
+use OCA\Mail\Db\MailAccountMapper;
 use OCA\Mail\Http\AttachmentDownloadResponse;
 use OCA\Mail\Http\HtmlResponse;
 use OCA\Mail\Service\ContactsIntegration;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\AppFramework\Http\ContentSecurityPolicy;
 use OCP\AppFramework\Http\JSONResponse;
 use OCP\AppFramework\Http\TemplateResponse;
 use OCP\IL10N;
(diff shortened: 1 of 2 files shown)


jancborchardt:
@stp-ip @RandolfCarter @LEDfan @Gomez please also test! :)

jancborchardt:
:+1: solves https://github.com/owncloud/mail/issues/206 for me!

Gomez:
Wow! Nice, works now. Thanks!
:beer:

jancborchardt:
@Gomez also thumbs-up https://github.com/owncloud/core/pull/16463 ;)

jancborchardt:
Just one thing: Without the core pull, HTML mails don’t load anymore:
```
{"reqId":"cHvovl3tzvf2ppWkFq7a","remoteAddr":"127.0.0.1","app":"PHP","message":"Call to undefined method OCP\\AppFramework\\Http\\ContentSecurityPolicy::disallowScriptDomain() at \/home\/jan\/owncloud\/apps\/mail\/lib\/controller\/messagescontroller.php#176","level":3,"time":"2015-05-20T11:06:26+00:00","method":"GET","url":"\/owncloud\/index.php\/apps\/mail\/accounts\/16\/folders\/SU5CT1g=\/messages\/10235\/html?requesttoken=2zqv5dD5\/L2Py6h\/kJiAbZS\/ER%2BAPX"}
```

Gomez:
Missed that. I get the same as @jancborchardt

LukasReschke:
> Missed that. I get the same as @jancborchardt

With this PR on master it will require the newest pull, older versions as in: ownCloud is not master should not require this.

callahad:
Ah, so the issue was the iframe sandbox property?

LukasReschke:
> Ah, so the issue was the iframe sandbox property?

Yes. When embedding a sandboxed iframe in Firefox and disallowing script execution any link opened from that iframe (also when having a blank target) will have the same limitations. - This is different than the behaviour of other browsers such as Chrome, Safari, etc.

As with oC 8.1 we can set a CSP policy on controller base we are now just forbidding any script execution from the controller itself using CSP and removed the iframe sandbox.

LukasReschke:
That said, I still believe this is a Firefox bug ;-)

PR #697, opened 2015-05-20
Fix iconv issue and allow proper display of some html messages

DeepDiver1975:
I had issues with displaying of email from codacy because of some character in the html body which could not be converted to utf-8 .

unit test included

@LukasReschke @irgendwie @zinks- @wurstchristoph please review - THX

### DIFF
--- a/lib/message.php
+++ b/lib/message.php
@@ -398,8 +398,18 @@ class Message {
 		return $data;
 	}
 
+	/**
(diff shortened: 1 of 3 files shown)


LukasReschke:
Makes sense :+1:

irgendwie:
:+1: should be fine!

PR #698, opened 2015-05-20
Change 'Save to Files' icon, add Archive icon

jancborchardt:
Please review @jviereck @sorglos @zinks- @biolumin

Before (looked like an »Upload« icon, thanks @jviereck for the feedback!):
capture du 2015-05-20 12 49 26
After:
capture du 2015-05-20 12 45 56

Archive icon:
capture du 2015-05-20 12 44 46


sorglos:
@jancborchardt: Both patches work as expected. Tested with oC 8.0.3.4 and 8.1.0.6. :+1:

jancborchardt:
@sorglos great, thank you for the review! You are also welcome in our IRC channel #owncloud-mail (on freenode) :)

biolumin:
What is the archive icon reference for?

jancborchardt:
@biolumin it’s an icon for the special folder »Archive« which exists for some servers if you archive stuff.

sorglos:
@biolumin: [IMAP LIST Extension for Special-Use Mailboxes](http://www.ietf.org/rfc/rfc6154.txt)
* \All
* \Archive
* \Drafts
* \Flagged
* \Junk
* \Sent
* \Trash

DeepDiver1975:
:+1:

biolumin:
I should have been more clear. I understand what and archive folder is, but why was it referenced for this job? It appears that this simple icon change had nothing to do with the archive folder or it's associated icon.

css/mail.css
```diff
@@ -602,6 +602,9 @@
 .icon-sent {
 	background-image: url('../img/sent.svg');
 }
+.icon-archive {
+	background-image: url('../img/archive.svg');
```
zinks-: @biolumin This merge is icon-related and adds the archive icon here (does this answer your question?)

biolumin: @zinks- Thanks for the clarification. I'm still familiarizing myself with the code, so I didn't realize that some things are integrated directly from core, like icons.css for example. It all makes sense now. :)

PR #703, opened 2015-05-20
'position: fixed' for new message button, fix #681

### LINKED ISSUE #681: Make 'New message' button static
If you have multiple accounts and/or accounts with many folders you often have to scroll down the folders list to get to the very last account. If you then want to compose a new mail, you have to scroll up again to be able to click the 'New message' button.

I think we should make the button static and scroll the folders list only

what do you think @jancborchardt?

jancborchardt:
Please review @wurstchristoph @Xenopathic @Henni @irgendwie :)

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -301,7 +301,16 @@
 }
 
 /* editor */
-
+#mail-new-message-fixed {
+	position: fixed !important;
+	width: 241px !important;
+	z-index: 1;
+	padding-bottom: 12px;
+	background-color: rgba(245, 245, 245, .75);
+}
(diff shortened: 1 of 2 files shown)


biolumin:
You'll want to create a new div for the message button above the unordered list. The way this is implemented may display incorrectly across browsers because scroll bars have varying widths depending on the OS, browser make, and browser model.

jancborchardt:
@biolumin how will a new div outside solve that? (While not simply overlapping the scrollbar.)

irgendwie:
Works as expected.

PR #709, opened 2015-05-21
Abort message loading when the composer is being shown

### LINKED ISSUE #708: Message is loaded/rendered although 'New message' button was clicked
I found another timing issue: When you click the 'New message' button while a message is loaded, the composer is shown shortly and then replaced by the message.

Steps to reproduce:
- click a message in the message list
- while the message loads (spinner visible), click the 'New message' button
- the composer is shown shortly, but then replaced by the clicked message

Expected behavior:
The message load request is aborted and the composer view is shown

wurstchristoph:
fixes #708

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -549,6 +549,14 @@ var Mail = {
 		openComposer: function(data) {
 			$('#mail_new_message').prop('disabled', true);
 
+			// Abort message loads
+			if (Mail.State.messageLoading !== null) {
+				Mail.State.messageLoading.abort();
+				$('iframe').parent().removeClass('icon-loading');
+				$('#mail-message').removeClass('icon-loading');
+				$('#mail_message').removeClass('icon-loading');
+			}
+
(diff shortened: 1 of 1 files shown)


irgendwie:
Problem:

If you click on a message (doesn't matter if it's text or html) and quickly click the "New message" Button, the spinner is **not** removed.
Meaning, you click the "New message" Button, before even the layout is loaded.

wurstchristoph:
@irgendwie It _should_ be fixed now

wurstchristoph:
@irgendwie @jancborchardt 
I updated the commit and fixed another issue: when clicking through messages fast, they are all rendered after each other when the request finishes. Previous message loading request should be aborted

jancborchardt:
:+1: looks good to me. :)

jancborchardt:
@irgendwie @Henni @colmoneill we need another review here. :)

colmoneill:
Yep this works great. :+1:

PR #711, opened 2015-05-21
add 'sentmail' to special folders detection

jancborchardt:
As talked about in https://github.com/owncloud/mail/issues/411#issuecomment-103142914, please review @MeCias @DeepDiver1975 @zinks-

### DIFF
--- a/lib/mailbox.php
+++ b/lib/mailbox.php
@@ -37,7 +37,7 @@ class Mailbox {
 	private $displayName;
 
 	/**
-	 * @var string 
+	 * @var string
 	 */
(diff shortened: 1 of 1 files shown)


MeCias:
No complains from my side, thank you for the work. :thumbsup:

zinks-:
:+1:

zinks-:
@jancborchardt  We should think about a setting, because these lists are going to grow!

jancborchardt:
@zinks- them growing is fine though. I don’t really see the need for a setting as we should fix this in the app for everyone. :)

PR #715, opened 2015-05-23
client-side message cache

wurstchristoph:
- [x] fetch first 10 messages of the active folder in background and cache them
- [x] additionally, cache 10 most recently loaded messages
- [ ] ~~cache HTML messages (HTML content)~~
- [x] automatically remove messages of deleted accounts
- [ ] automatically fetch new messages loaded in background

partially implements #480

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -1,4 +1,4 @@
-/* global Handlebars, Marionette, Notification, relative_modified_date, formatDate, humanFileSize, views, OC */
+/* global Handlebars, Marionette, Notification, relative_modified_date, formatDate, humanFileSize, views, OC, _ */
 
 if (_.isUndefined(OC.Notification.showTemporary)) {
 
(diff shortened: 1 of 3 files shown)


wurstchristoph:
@LukasReschke is it save to cache the content of HTML messages (iframe)?

If the message is loaded for the first time, I would store the HTML content of the iframe to local storage. The next time it is loaded, I would load this HTML into the iframe directly.

LukasReschke:
This approach will basically render our in-depth CSP protection useless. Our current security model relies on:

1. The HTML sanitizer which tries to sanitize the HTML
2. The sanitized HTML is then put in a CSP protected iframe

With this change we're left with No. 1 as the CSP protection basically goes away. The reason behind our CSP protection is that we pretty much don't trust our HTML sanitizer because it's pretty easy to do things wrong there, that would have catastrophic consequences. With CSP an attacker can do nothing at all :-)

You can test this by modifying your local storage to have a `<script>alert(1)</script>`, you will see that the script gets executed. With our iframed HTML mails even JS like that doesn't get execute at all.

I thought about using an sandboxed iframe here and inject the data into the iframe. This would however not be possible without the same origin attribute which renders this pretty useless again. Seems like we need to come up with a workaround for CSP here as well.

What I can imagine is that it might be possible to have dummy endpoint that just serves our CSP policy and we load this endpoint (we can cache the endpoint using the browser HTTP cache by setting some caching headers) and then insert our HTML into this. Then it *may* be the chance that CSP also applies here :pray: – Let's dig deeper :beers:

irgendwie:
tried to produce some bugs but seems to work pretty good, merge-able! :+1:

Is it intended to only cache messages, which have been opened recently?
I thought of caching the 20 latest messages which have been received,
to speed up working with the latest received messages?

wurstchristoph:
@irgendwie thank you for reviewing! :)

Yes, this is intended. However, this is just 1 out of many strategies of choosing which messages to cache ;-)

irgendwie:
Yeah, thought so ;-) thanks for clarifying!

jancborchardt:
Similar to @irgendwie I'd also expext the last 20 _received_ messages to be cached. That's more important than caching messages which have already been opened I'd say. :)

wurstchristoph:
@irgendwie @jancborchardt I updated my PR (see description above)
I still like the idea of additionally caching most recently loaded messages as it speeds up switching back and forth between two or more messages even if they are not on the top :rocket: 

I also refactored where I thought it's needed: data management (fetching/caching) is now done in Mail.Communication. Therefore Mail.UI does only handle UI related stuff (that's what it really should do, nothing more) :wrench:

irgendwie:
I approve the newest change in your PR! :+1: Speeds things up for my working behavior.

Still I think you're right, that caching a few recently loaded messages would also speed things up for other working behaviors, so why not try a hybrid approach and cache the newest 20 messages **and** most recently used messages.

wurstchristoph:
@irgendwie thats' exactly what this PR does: cache 10 newest __and__ 10 most recently loaded messages ;-)

irgendwie:
Awesome! :-) I totally approve this!

Am 30. Mai 2015 12:25:04 MESZ, schrieb Christoph Wurst <notifications@github.com>:

jancborchardt:
Approving this behavior as well! :) - cant test cause vacation :^)

wurstchristoph:
> codacy/pr — Not so good... This commit quality could be better. 

What's the problem there? When I view the details, there's nothing listed. Can I ignore that @DeepDiver1975?

wurstchristoph:
I'll implement "automatically fetch new messages loaded in background" in another PR as this one is rather big. Let's do it step by step :)

jancborchardt:
So merge? Or last rwview @Deepdiver1975

jancborchardt:
@DeepDiver1975 @zinks- @AidanAmavi another review? :)

aidanamavi:
:+1: Works well.

DeepDiver1975:
Wow - impressive!  Thanks. 

I have only one issue: all message loading code should live in the backbone collection for messages (somehow).

This change is an improvement for now but we might want to move/change it in the future. 

:+1:

PR #736, opened 2015-05-27
prefill cc from draft if set

### LINKED ISSUE #619: cc is not filling in from draft
#613 tried to fix cc problem but it didn't fix the whole thing.
I think `value="{{replyCc}}"` was introduced from filling in cc from draft, but the problem is that `replyCc` is not a list. I think to fix it down to the bottom we need to make draft in local storage to store CC in list, then use is directly, or parse what is stored to make it a list, then fill it in.

wurstchristoph:
fixes #619

@clementhk

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -292,6 +292,22 @@ var Mail = {
 				return str;
 			});
 
+			Handlebars.registerHelper("ifHasCC", function (cc, ccList, options) {
+				if (!_.isUndefined(cc) || !_.isUndefined(ccList)) {
+					return options.fn(this);
+				} else {
+					return options.inverse(this);
+				}
+			});
+
+			Handlebars.registerHelper("unlessHasCC", function (cc, ccList, options) {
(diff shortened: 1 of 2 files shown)


clementhk:
Ok I would prefer to cache cc as list so it uses the method to print, but this should work.

Review-by: me.

js/mail.js
```diff
@@ -292,6 +292,22 @@ var Mail = {
 				return str;
 			});
 
+			Handlebars.registerHelper("ifHasCC", function (cc, ccList, options) {
+				if ((cc !== undefined) || (ccList !== undefined)) {
```
DeepDiver1975: I prefer the use of _.isUndefined() - compare more reliably

wurstchristoph:
@DeepDiver1975 updated my PR. Please review again ;-)

jancborchardt:
@clementhk is this now cool from your POV as well? And @DeepDiver1975?

DeepDiver1975:
:+1:

PR #737, opened 2015-05-29
Introduce php mail as alternative transport

### LINKED ISSUE #729: Support sending mail through php's builtin mail() function
Some webhosts disable outgoing connections on common smtp ports. They often allow sending mail throught the php mail() function, which connects to some smtp server chosen by the hoster.

Web hosting services known to block smtp connections :

- 1&1 (1und1)

DeepDiver1975:
fixes #729

How to enable:
- add app.mail.transport with value php-mail to your config.php

### DIFF
--- a/lib/service/logger.php
+++ b/lib/service/logger.php
@@ -14,18 +14,12 @@ namespace OCA\Mail\Service;
 
 use OCP\ILogger;
 
-class Logger {
+class Logger implements ILogger {
 
-	/**
-	 *
-	 * @var array
-	 */
+	/** @var array */
 	private $context;
 
-	/**
-	 *
-	 * @var ILogger
-	 */
+	/** @var ILogger */
 	private $logger;
 
 	/**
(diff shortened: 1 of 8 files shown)


DeepDiver1975:
@zinks- @marius01 feel free to test this

marius01:
Will check this. Thank you.

jancborchardt:
@zinks- @marius01 did you test this now? Would be cool if you can, especially since you were the ones who requested it ;)

marius01:
Hello.

It does not work. 
Creating account failed: Mail server denied authentication.

It does not work, if i type something in the SMTP-Boxes and still not, if i type nothing in it :(

DeepDiver1975:
> Sorry for the late answer. I'm sorry having to tell that it does not work.
> 
> Erstellung des Kontos ist fehlgeschlagen:Error connecting to SMTP server.
> 
> It does not work, if i type something in the SMTP-TextBox and still not, if i type nothing in it :(

hmm ... this is honestly strange - are you sure you did apply the patch of this PR properly? THX

DeepDiver1975:
Ok - just tested it once again - works for me *now*

Gomez:
Requirement is that php in ownClouds E-Mail-Server settings works? Right?

DeepDiver1975:
> Requirement is that php in ownClouds E-Mail-Server settings works? Right?

No - we are using php's mail function http://php.net/manual/de/function.mail.php - from my understanding the proper settings have to go into php.ini.

DeepDiver1975:
> Requirement is that php in ownClouds E-Mail-Server settings works? Right?

But you are right - why not simply reuse core's email configuration?

Should we add this as another option or instead of php-mail?

Gomez:
I would say instead of php-mail, the ownClouds E-Mail-Server setting offer php sending too? Just reuse the selected option there. 
As a bonus we get the "Test email settings" on top.

DeepDiver1975:
> I would say instead of php-mail, the ownClouds E-Mail-Server setting offer php sending too? Just reuse the selected option there.
> As a bonus we get the "Test email settings" on top.

agreed - will change this then ...

DeepDiver1975:
> agreed - will change this then ...

hmmm ... the API of ownCloud's IMailer is not really compliant *yet* to allow this - let's addess this in a later PR.

zinks-:
And allowing users to send mail using these settings should be set only by admins

jancborchardt:
@Gomez review plz :)

jancborchardt:
@zinks- @ghost please also review this again! :)

Gomez:
I get this error:
```
302 {"reqId":"mmCYlubYK7uTPJSBH\/KH","remoteAddr":"127.0.0.1","app":"PHP","message":"mail() expects parameter 1 to be string, array given at \/var\/www\/mail_dev\/core\/apps\/mail\/vendor\/pear-pear.horde.org\/H    orde_Mail\/Horde\/Mail\/Transport\/Mail.php#103","level":3,"time":"2015-06-17T19:51:03+00:00"}
```
Can some1 confirm, i have a upstream bug. Just not sure, because only me with this error?
@jancborchardt @DeepDiver1975 @zinks-

DeepDiver1975:
@Gomez yes - I'm observing the same issue - we shall fix this upstream

Gomez:
Upstream PR:
https://github.com/horde/horde/pull/138

Gomez:
@DeepDiver1975 @zinks- This is ready now. 
Could some with a 1&1 (1und1) account test?

Gomez:
As this is a (optional) switch and the original reporter is gone (deleted user?) we should merge it and handle reports in new issues?

Gomez:
will rebase

DeepDiver1975:
:+1:

lib/service/logger.php
```diff
@@ -41,67 +35,65 @@ public function __construct($appName, ILogger $logger) {
 	}
 
 	/**
-	 * 
-	 * @param string $message
+	 * @inheritdoc
 	 */
-	public function emergency($message) {
+	public function emergency($message, array $context = array()) {
```
clementhk: Sorry to ask this late, but what are these `array $context = array()` for?

DeepDiver1975:
the context is an array which holds the variables to be used for string interpolation

```php

$logger->emergency('Hello, {user}!', ['user' => 'clementhk']);

DeepDiver1975:
Same logic as with PSR Logger

Issue #614, opened 2015-05-05
'Delete permanently' tipsy not shown for next message


wurstchristoph:
The tipsy should be shown for the next message in the message list if one clicks the trash icon.

Steps to reproduce
- go to trash folder
- click  trash icon of a message (not the first one)
- watch how it vanishes and the next one comes up
- the next messages tipsy should be shown, even if you don't move your cursor

There is no issue when deleting the first message in the list, the next tipsy is shown immediately.

tipsy was introduced here https://github.com/owncloud/mail/pull/574

jancborchardt:
@irgendwie can you check that out? :)

PR #741, opened 2015-06-01
Fix tipsy not shown for next message in trash

irgendwie:
Fix for bug #614.
Seems like it's a jQuery bug?!

@wurstchristoph @jancborchardt review please!

### DIFF
--- a/js/views/message.js
+++ b/js/views/message.js
@@ -105,6 +105,9 @@ views.Message = Backbone.Marionette.ItemView.extend({
 					Mail.UI.openMessage(nextMessage.id);
 				}
 			}
+			// manually trigger mouseover event for current mouse position
+			// in order to create a tipsy for the next message if needed
+			$(document.elementFromPoint(event.clientX, event.clientY)).trigger('mouseover');
 		});
 
 		// really delete the message


irgendwie:
Ok, doesn't seem like a jQuery bug and more like the tipsy is generated, but at the wrong location. (Probably a compatibility issue between the tipsies and jQuery)

After the slideup function call, we delete every tipsy, which leads to no tipsy at all for the next message, because the mouseover event was already called.
I solved this by manually calling the mouseover event for the current mouse position, which leads to a correct placed tipsy for the next message in the list (and only if the mouse is over the element of course).

Gomez:
Is this still valid? I dont  get "Delete permanently" at all in trash?

irgendwie:
@Gomez There should be tipsies in the trash! For Firefox and Chrome, they are visible with the current master for me!

Gomez:
@irgendwie Hmm, would you send me a screenshot?

irgendwie:
@Gomez 

tipsy

jancborchardt:
Fixes #614 :+1: good stuff! :)

jancborchardt:
(And yeah, I do get tipsies in the Trash as intended.)

Gomez:
No tipsies for me. And the original-title text is missing. Maybe something with special folders?

```
<div original-title="" class="icon-delete action delete"></div>
```

jancborchardt:
original-title is getting filled in by Tipsy normally, from the »title« attribute. Any JS errors? Does it work with another account or in another browser?

wurstchristoph:
:+1: fixes it!

Gomez:
@jancborchardt No Tipsy for me :( Let us look at the next meetup...

jancborchardt:
@Gomez you and you strange custom setups ;)

Gomez:
@jancborchardt You and your average/simple/boring setups :)

PR #744, opened 2015-06-05
JSCS for send-mail.js

Gomez:
One more file.
209 errors to go

### DIFF
--- a/js/send-mail.js
+++ b/js/send-mail.js
@@ -1,5 +1,5 @@
 /* global Mail, OC */
-$(function () {
+$(function() {
 
 	function split(val) {
 		return val.split(/,\s*/);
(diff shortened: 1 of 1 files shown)


Gomez:
@DeepDiver1975 @wurstchristoph @clementhk @Xenopathic Easy review..

js/send-mail.js
```diff
@@ -151,14 +151,16 @@ $(function () {
...
+		newMessageCcBcc.find('#cc').focus();
 		$('#new-message-cc-bcc-toggle').fadeOut();
 	});
 
 	$(document).on('click', '#reply-message-cc-bcc-toggle', function() {
-		$('#reply-message-cc-bcc').slideToggle();
-		$('#reply-message-cc-bcc #cc').focus();
+		var replyMessageCcBcc = $('#reply-message-cc-bcc');
+		replyMessageCcBcc.slideToggle();
+		replyMessageCcBcc.find('#cc').focus();
```
wurstchristoph: Is this replacement really the same? I may be wrong, but to me it looks like it was a selection if any element with classes ``reply-message-cc-bcc`` and ``cc`` before.
Your changes select any element of class ``reply-message-cc-bcc``, but then ``find`` searches any descendants of that selection, see https://api.jquery.com/find/

Gomez: Possible, i will check it in a other pr. Reverted it here, as it has nothing todo with jscs. THX

wurstchristoph:
:+1:

Issue #664, opened 2015-05-12
It loads all the emails rather than the 20-30 it's supposed to.


Azelphur:
the mail app tries to load all emails for me rather than the 20-30 it supposed to, this results in a rather long hang followed by it being unable to load my inbox as it has >8,000 emails in it.

Here's my horde.log

````
------------------------------
>> Tue, 12 May 2015 00:09:46 +0000
>> Connection to: imap://imap.gmail.com:993/
>> Server connection took 0.0824 seconds.
S: * OK Gimap ready for requests from 2001:4d48:ad52:4800:2a92:4aff:fe34:56f1 m138mb130605605wma
C: 1 CAPABILITY
S: * CAPABILITY IMAP4rev1 UNSELECT IDLE NAMESPACE QUOTA ID XLIST CHILDREN X-GM-EXT-1 XYZZY SASL-IR AUTH=XOAUTH2 AUTH=PLAIN AUTH=PLAIN-CLIENTTOKEN
S: 1 OK Thats all she wrote! m138mb130605605wma
>> Command 1 took 0.0209 seconds.
C: 2 AUTHENTICATE PLAIN [INITIAL CLIENT RESPONSE (username: azelphur@azelphur.com)]
S: * CAPABILITY IMAP4rev1 UNSELECT IDLE NAMESPACE QUOTA ID XLIST CHILDREN X-GM-EXT-1 UIDPLUS COMPRESS=DEFLATE ENABLE MOVE CONDSTORE ESEARCH UTF8=ACCEPT
S: 2 OK azelphur@azelphur.com authenticated (Success)
>> Command 2 took 0.3945 seconds.
C: 3 ENABLE UTF8=ACCEPT
C: 4 LIST "" *
S: * ENABLED UTF8=ACCEPT
S: 3 OK Success
>> Command 3 took 0.1213 seconds.
S: * LIST (\HasNoChildren) "/" "Drafts"
S: * LIST (\HasNoChildren) "/" "INBOX"
S: * LIST (\HasNoChildren) "/" "Sent"
S: * LIST (\HasChildren) "/" "Trash"
S: * LIST (\HasNoChildren) "/" "Trash/Follow up"
S: * LIST (\HasNoChildren) "/" "Trash/Misc"
S: * LIST (\HasNoChildren) "/" "Trash/Priority"
S: * LIST (\Noselect \HasChildren) "/" "[Gmail]"
S: * LIST (\HasNoChildren \All) "/" "[Gmail]/All Mail"
[...]

Issue #533, opened 2015-04-24
Loading messages is performance-heavy


Xenopathic:
It took my Raspberry Pi 128 seconds to load my inbox. Not good. The whole time the CPU was pegged at 100% too. Using IMAPS going through Gmail (my inbox is pretty big though, 100s of mails).

jancborchardt:
It only loads the first 20 messages though, so a big inbox shouldn’t really be slower.

But … Raspberry Pi ;)

Xenopathic:
@jancborchardt Yeah, but for only the first 20 messages over 2 minutes is a *little* excessive...

jancborchardt:
Yup, that’s true.

Xenopathic:
I'll do some Blackfire benchmarks at some point, after 0.1 perhaps.

jancborchardt:
@Xenopathic show me the benchmarks ;D

ChristophWurst:
The performance should be better since https://github.com/owncloud/mail/pull/1486

PR #747, opened 2015-06-06
We shall not search in case we only want to get all messages

DeepDiver1975:
refs #664 
refs #533 

Please test @Azelphur @clementhk @wurstchristoph @irgendwie @Xenopathic 

Please note searching itself will remain unchanged and bad performance is still expected.

### DIFF
--- a/lib/mailbox.php
+++ b/lib/mailbox.php
@@ -89,23 +89,31 @@ class Mailbox implements IMailBox {
 	}
 
 	public function getMessages($from = 0, $count = 2, $filter = '') {
-		if ($filter instanceof Horde_Imap_Client_Search_Query) {
-			$query = $filter;
+
+		if (empty($filter)) {
+			$total = $this->getTotalMessages();
+			$from = $total - $from;
+			$to = max($from - $count + 1, 1);
+			$ids = new \Horde_Imap_Client_Ids("$from:$to", true);
 		} else {
-			$query = new Horde_Imap_Client_Search_Query();
-			if ($filter) {
-				$query->text($filter, false);
+			if ($filter instanceof Horde_Imap_Client_Search_Query) {
+				$query = $filter;
+			} else {
+				$query = new Horde_Imap_Client_Search_Query();
+				if ($filter) {
+					$query->text($filter, false);
+				}
 			}
+			if ($this->getSpecialRole() !== 'trash') {
+				$query->flag(Horde_Imap_Client::FLAG_DELETED, false);
+			}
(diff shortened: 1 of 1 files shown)


irgendwie:
> C: 6 FETCH 4094:4114 (ENVELOPE FLAGS RFC822.SIZE UID INTERNALDATE BODYSTRUCTURE BODY.PEEK[HEADER.FIELDS (IMPORTANCE LIST-POST X-PRIORITY CONTENT-TYPE)])
S: 6 OK Success
-- Command 6 took 0.1766 seconds.
C: 7 LOGOUT
S: * BYE LOGOUT Requested
S: 7 OK 73 good day (Success)
-- Command 7 took 0.1807 seconds.

Seems like the fetch doesn't get any messages for my big folder (~4000 mails in it).
Still works for folders with less mails in it (tested with folders containing up to 90 mails).

Any idea @DeepDiver1975?

DeepDiver1975:
Grrrr .... let me have a look.

DeepDiver1975:
@irgendwie can I ask you to send a status call to the server for that big folder? Would be interesting to see the response - THX

jancborchardt:
@Xenopathic please test this as it tackles an issue you reported.

@irgendwie more details please. :)

jancborchardt:
@Azelphur please also test this as it tackles your issue https://github.com/owncloud/mail/issues/664

Gomez:
@DeepDiver1975 Whats the plan here?

ChristophWurst:
> searching itself will remain unchanged and bad performance is still expected

Not necessarily, it's actually about twice as fast with my email provider.

lib/mailbox.php
```diff
@@ -89,23 +89,31 @@ public function __construct($conn, $mailBox, $attributes, $delimiter='/') {
...
 
 	public function getMessages($from = 0, $count = 2, $filter = '') {
-		if ($filter instanceof Horde_Imap_Client_Search_Query) {
-			$query = $filter;
+
+		if (empty($filter)) {
+			$total = $this->getTotalMessages();
+			$from = $total - $from;
+			$to = max($from - $count + 1, 1);
+			$ids = new \Horde_Imap_Client_Ids("$from:$to", true);
```
ChristophWurst: This won't work in all cases as sorting is not guaranteed AFAIK

ChristophWurst:
This sounds like a reasonable solution:
http://stackoverflow.com/questions/1387999/is-it-possible-to-use-imap-paging
http://stackoverflow.com/questions/1387999/is-it-possible-to-use-imap-paging#comment1228442_1388012

Gomez:
Is this fixed with #1486

ChristophWurst:
yes, #1486 replaced this PR more or less

PR #748, opened 2015-06-06
[WIP] Threaded message view

DeepDiver1975:
refs #21

Conversations/threads will be tough to implement.

TODO:
- [ ] unread flag has to be calculated based on all thread members
- [ ] no more mark-as-read on simple click
- [ ] change detection has to be redone - e.g. logic of adding new messages
- [ ] elaborate on how to get the full conversation for each loaded message 
- [ ] how does reply/forward fit into the threaded view
- [ ] ....

I'd say this is a complete topic on it's own -> 0.3

@jancborchardt

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -266,6 +266,13 @@
 	margin-left: 37px;
 }
 
+.mail_message_summary_thread_count {
+	display: initial;
+	position: absolute;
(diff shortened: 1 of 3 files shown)


Gomez:
Outdated, lets discuss further in #21

PR #752, opened 2015-06-10
fix reply fields overlapping message header bar

jancborchardt:
Previously when you had a large reply box and scrolled down, the reply input boxes would overlay the fixed header bar. This fixes that.

Please review @aidanamavi @colmoneill @irgendwie :)

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -496,6 +496,7 @@
 	height: 90px;
 	width: 50%;
 	background-color: rgba(255,255,255,.9);
+	z-index: 100;
 }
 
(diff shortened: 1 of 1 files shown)


colmoneill:
Just to be sure: we're talking about this &darr;

[Imgur](http://i.imgur.com/7080mVX.png)

If this is the issue, then yep, confirmed, the z-index fixes it!

:+1:

jancborchardt:
Yup, that’s it! A second review please @aidanamavi @irgendwie @zinks-

PR #754, opened 2015-06-10
return HTTP 404 if message does not exist

wurstchristoph:
without this change, the application returns HTTP 500 and an error page
("This email (808) can't be found. Probably it was deleted from the server recently. Please reload.")


DeepDiver1975:
:+1:

PR #755, opened 2015-06-10
make Mail app work on mobile again, fix #457

### LINKED ISSUE #457: Make interface work on mobile again
The one problem we have with the 3-column layout now is that it doesn’t properly work on mobile anymore. The sidebar correctly hides to the left, but both the message list and the mail content is displayed next to each other.

Maybe on mobile we need to have the content view show the message list, and then overlay the message when one is clicked (with a close button on the top right). Similar to how the pictures app works.

What do you think @owncloud/designers @DeepDiver1975 @wurstchristoph @Gomez @zinks-?

jancborchardt:
Please review @owncloud/designers @DeepDiver1975 @wurstchristoph @aidanamavi @colmoneill 

This is only the first pass, making the view work on mobile. What has to be done in follow-up PRs is to also show the actions like delete and favorite in the detail view.

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -507,6 +507,10 @@
 	margin-bottom: 0;
 }
 
+#mail-message-close {
+	display: none;
+}
+
 .mail-message-body {
 	margin: 100px 10px 50px 30px;
 }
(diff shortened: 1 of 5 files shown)


DeepDiver1975:
Nice :+1:

wurstchristoph:
Awesome, looks very good :+1:

Issue #753, opened 2015-06-10
Deletion broken, does not switch to other message and current message still stays in detail view


jancborchardt:
Just test it with current master cc @DeepDiver1975 @zinks- @irgendwie @AidanAmavi 

Correct behavior should be: https://github.com/owncloud/mail/issues/730

irgendwie:
Werde es mir in den nächsten Tagen mal ansehen!

irgendwie:
Fixed by PR #757

jancborchardt:
@irgendwie hehe, @aidanamavi was faster than you. ;)

irgendwie:
:see_no_evil: 

![I'm not even mad](http://www.mememaker.net/static/images/memes/3153718.jpg)

jancborchardt:
lol, good one

aidanamavi:
lol

PR #757, opened 2015-06-11
Fixed message deletion bug, and message loading.

### LINKED ISSUE #730: On deletion, switch to newer mail instead of older
Currently when you delete an email, the next one automatically getting opened is the next older one. Usually email apps do it the other way around: Opening the next newer one.

We should switch that around hence. :) (That also makes it simpler to follow conversations for now while will don't have threading #21 - simply start at the oldest one and delete to get to the next one.)
Of course when you delete the first message in the current folder there should be a special case where the second message (the older one) is opened instead.

aidanamavi:
Fixed delete function.
Fixed message loading function.
Modified delete function to load a newer message upon deletion, otherwise load an older message upon deletion.

Resolves issues #730 and #753

### DIFF
--- a/js/views/message.js
+++ b/js/views/message.js
@@ -98,14 +98,14 @@ views.Message = Backbone.Marionette.ItemView.extend({
 			$('.tipsy').remove();
 			var thisModelCollection = thisModel.collection;
 			var index = thisModelCollection.indexOf(thisModel);
-			var nextMessage = thisModelCollection.at(index+1);
+			var nextMessage = thisModelCollection.at(index-1);
 			if (!nextMessage) {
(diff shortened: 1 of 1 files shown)


jancborchardt:
Nice, works great and fixes both issues! :+1: another review please @colmoneill @irgendwie @wurstchristoph

wurstchristoph:
Nice :+1:

jancborchardt:
Awesome @aidanamavi! Good to have you on board as JS wiz. :)

aidanamavi:
Thanks @jancborchardt! I'm happy I can contribute in some way :)

PR #760, opened 2015-06-11
resolved message fetcher timing issues

wurstchristoph:
The background message fetcher didn't stop fetching messages of the previous
folder when the folder was changed. This caused 404s because the message
IDs and folder IDs did not match

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -326,10 +326,8 @@ var Mail = {
 					// Stop previous fetcher
 					clearInterval(timer);
 
-					// Stop any ajax message requests
-					if (Mail.State.messageLoading !== null) {
-						Mail.State.messageLoading.abort();
-					}
+					// Clear waiting queue
+					queue.length = 0;
 
 					// Start again
 					this.start();


jancborchardt:
How can the issue be tested/reproduced, so we can test the fix? :)

wurstchristoph:
@jancborchardt the steps are as follows:
1. clear the cache via ``localStorage.clear()``
2. open the mail app
3. open a folder (e.g. inbox)
4. while messages are fetched (see the browser console or similar) switch to another folder
5. there could be some 404 errors :)

jancborchardt:
Ok, when I tested it with current master and clicking around through some folders, there were 404s sometimes. When I tested with this branch those did not occur. So I guess :+1:  ;)

wurstchristoph:
@jancborchardt great, that's exactly what should be resolved by this PR ;-)

wurstchristoph:
@aidanamavi @irgendwie please review :)

aidanamavi:
:+1: Works well. Great job @wurstchristoph :)

PR #763, opened 2015-06-11
[WIP] use horde sync to keep messages and message lists up to date

### LINKED ISSUE #733: Detect flag changes and expunged messages
ATM we only check for new messages (uid validity, next uid). Additionally, it would be great to detect and sync any other changed flags (unseen, starred, replied, deleted,...) or messages deleted in other IMAP clients.

:books:  https://tools.ietf.org/html/rfc4549#section-4.3 seems to be a good strategy to do that

Related: #663

ChristophWurst:
fixes #733

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -619,6 +619,9 @@ var Mail = {
 					OC.generateUrl('apps/mail/accounts/{accountId}/folders/{folderId}/messages',
 						{'accountId': accountId, 'folderId': folderId}), {
 						success: function (jsondata) {
+							var messages = jsondata.messages;
+							var syncToken = jsondata.syncToken;
+							console.log('sync token: ' + syncToken);
 							Mail.State.currentlyLoading = null;
 							Mail.State.currentAccountId = accountId;
 							Mail.State.currentFolderId = folderId;
@@ -628,19 +631,19 @@ var Mail = {
 							// Fade out the message composer
 							$('#mail_new_message').prop('disabled', false);
 
-							if (jsondata.length > 0) {
-								Mail.UI.addMessages(jsondata);
+							if (messages.length > 0) {
+								Mail.UI.addMessages(messages);
 
 								// Fetch first 10 messages in background
-								_.each(jsondata.slice(0, 10), function (message) {
+								_.each(messages.slice(0, 10), function (message) {
 									Mail.BackGround.messageFetcher.push(message.id);
 								});
 
-								var messageId = jsondata[0].id;
+								var messageId = messages[0].id;
 								Mail.UI.loadMessage(messageId);
 								// Show 'Load More' button if there are
 								// more messages than the pagination limit
-								if (jsondata.length > 20) {
+								if (messages.length > 20) {
 									$('#load-more-mail-messages')
 										.fadeIn()
 										.css('display', 'block');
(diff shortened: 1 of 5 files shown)


DeepDiver1975:
Nice start - thx. Just perfect timing - I need this fir the unified inbox as well.

js/mail.js
```diff
@@ -119,6 +119,38 @@ var Mail = {
...
+		var syncTokens = {};
+		// Update every 15 seconds
+		var updateInterval = 15;
+		var timer = null;
+
+		function detectChanges() {
+			var accountId = Mail.State.currentAccountId;
+			var folderId = Mail.State.currentFolderId;
+			// Only start if there ary any accounts/folders/tokens
+			if (accountId && folderId && syncTokens > 0) {
```
wurstchristoph: TODO: syncTokens.length

wurstchristoph:
@DeepDiver1975 I'll need your help on this. I always get an exception [here](https://github.com/horde/horde/blob/master/framework/Imap_Client/lib/Horde/Imap/Client/Data/Sync.php#L157).

Are there any preconditions to make sync work? There's very little documentation on that :/

DeepDiver1975:
I will debug into this the next days. Promised.

wurstchristoph:
Some further trial and error shows that sync __does__ work when used right after requesting the token. I could successfully detect any changes [here](https://github.com/owncloud/mail/pull/763/files#diff-ee701d435b416eddecc1798126969f4dR125).

Moreover I could find some code using the sync functionality: https://github.com/horde/horde/blob/3aff0b0dfcfb190e307b3dc01edfa87d4baaacc9/imp/lib/Ajax/Application/ListMessages.php#L384

I'm not totally sure, but it could be that we need server-side caching (I assume horde needs to know the previous mailbox state to detect any changes). I enabled caching in my dev env using `` 'app.mail.server-side-cache.enabled' => true``, but this didn't help neither :(

appinfo/routes.php
```diff
@@ -32,6 +32,10 @@
 				'url' => '/accounts/{accountId}/folders/{folderId}/messages/{messageId}/toggleStar',
 				'verb' => 'POST'],
 			[
+				'name' => 'messages#detectChanges',
```
DeepDiver1975: why is there another route? I don't get this - sorry

lib/controller/messagescontroller.php
```diff
@@ -332,4 +362,14 @@ private function buildHtmlBodyUrl($accountId, $folderId, $id) {
...
 	}
 
+	/**
+	 * 
+	 * @param \OCA\Mail\Mailbox $mailBox
+	 */
+	private function getSyncToken(Mailbox $mailBox) {
+		$m = new Account($this->getAccount());
+		$client = $m->getImapConnection();
+		return $client->getSyncToken($mailBox->getHordeMailBox());
```
DeepDiver1975: Oh - the sync token is mailbox based? Hmmmm ....

appinfo/routes.php
wurstchristoph: > why is there another route? I don't get this - sorry

Will be merged with the existing ``folders#detectChanges`` route, I added this just to try things out ;-)

PR #766, opened 2015-06-13
cache message lists (folders) in local storage

ChristophWurst:
partially implements #480 

IRC log:
> jancborchardt: christophwu: the only weird thing is that it still takes long switching folders (and accounts). Can’t we show the data from the cache?

@jancborchardt this PR should speed up switching folders ;-)

### DIFF
--- a/js/views/message.js
+++ b/js/views/message.js
@@ -273,30 +273,16 @@ views.Messages = Backbone.Marionette.CompositeView.extend({
 			.val(t('mail', 'Loading …'))
 			.prop('disabled', true);
 
-		var url = OC.generateUrl(
-			'apps/mail/accounts/{accountId}/folders/{folderId}/messages?from={from}&to={to}',
-			{
-				'accountId': Mail.State.currentAccountId,
-				'folderId':Mail.State.currentFolderId,
-				'from': from,
-				'to': from + 20
-			});
-		if (this.filterCriteria) {
-			url = OC.generateUrl(
-				'apps/mail/accounts/{accountId}/folders/{folderId}/messages?filter={query}&from={from}&to={to}',
-				{
-					'accountId': Mail.State.currentAccountId,
-					'folderId':Mail.State.currentFolderId,
-					'query': this.filterCriteria.text,
-					'from': from,
-					'to': from + 20
-				});
-		}
 		var self = this;
-		$.ajax(url, {
-				data: {},
-				type:'GET',
-				success: function (jsondata) {
+		Mail.Communication.fetchMessageList(
+			Mail.State.currentAccountId,
+			Mail.State.currentFolderId,
+			{
+				from: from,
+				to: from + 20,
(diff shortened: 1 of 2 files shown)


wurstchristoph:
I will update the background updating mechanism once PR #763 is finished, for now I simply trigger the folder reload event to fetch any changes.

jancborchardt:
There are some timing issues when clicking on other folders while the previous one is still loading. ;) Same story as in the past – best just cancel the previous request and replace the message list directly when a new folder is clicked.

wurstchristoph:
There are two timing issues as far as I can tell
- [x] requests triggered because you click a folder that is not cached yet
- [ ] requests triggered because a cached folder is clicked and the folder is being refreshed

To solve the second scenario there are two possible solutions:
- rework how folder lists get updates (is currently done inside the view?!)
- don't trigger folder refreshing when loading a cached message list (easy one)

I prefer the latter as I would like to include the horde sync stuff once it's ready which should replace folder updating anyway. For now this would mean if you switch folders, you would have to manually refresh it by clicking the button.
@jancborchardt  what do you think?
The first one would need another PR because this one is already pretty big. The second one could be done here (would actually make the PR smaller) ;-)

jancborchardt:
Hmm … needing to manually refresh the folder after switching to it sounds veeeery ugly. I would even say it’s better if we manually fire a click on the »Check for new messages« via JS after switching the folder.

wurstchristoph:
@jancborchardt alright, I updated the PR. Please review again, I hope I found and fixed all timing issues. Folders refresh is automatically triggered when you switch folders (regardless of the folder being cached or not) ;-)

jancborchardt:
Crazy shit! :+1: 

Another review please @irgendwie @DeepDiver1975 @aidanamavi @LukasReschke :)

jancborchardt:
@ChristophWurst needs rebase! :)

After that, a second review please homies!

ChristophWurst:
@jancborchardt done ;-)

ChristophWurst:
Rebased, so anybody can test it with the unified inbox :-)

js/views/message.js
```diff
@@ -277,30 +277,16 @@ views.Messages = Backbone.Marionette.CompositeView.extend({
...
-					'from': from,
-					'to': from + 20
-				});
-		}
 		var self = this;
-		$.ajax(url, {
-				data: {},
-				type:'GET',
-				success: function (jsondata) {
+		Mail.Communication.fetchMessageList(
```
DeepDiver1975: with respect to modularity and testability (one day - LOL) I guess we should inject the Communications object into the view ..... one day ...

jancborchardt:
So what’s the status here? Has been hanging around for 2 weeks now without new reviews. @DeepDiver1975 @Xenopathic @aidanamavi @irgendwie please check! :)

PR #768, opened 2015-06-14
load HTML messages as data source, so they can be cached

ChristophWurst:
@LukasReschke please review and tell me whether this creates a new security issue or if we can use this way of loading the iframe :smiley:

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -356,7 +356,13 @@ var Mail = {
 				}), {
 					data: {},
 					type: 'GET',
-					success: options.onSuccess,
+					success: function (message) {
+						if (message.hasHtmlBody) {
+							message.htmlBodyUrl = 'data:text/html;charset=utf-8,' +
+								encodeURI(message.body);
+						}
+						options.onSuccess(message);
+					},
 					error: options.onError
 				});
 			if (!options.backgroundMode) {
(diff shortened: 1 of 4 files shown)


LukasReschke:
Are you sure this works?

I had to set `$csp->addAllowedFrameDomain('data:');` in the pagecontroller, but at least Chrome will still throw a security exception:

```
Uncaught SecurityError: Failed to read the 'contentDocument' property from 'HTMLIFrameElement': Blocked a frame with origin "http://localhost" from accessing a frame with origin "null".  The frame requesting access has a protocol of "http", the frame being accessed has a protocol of "data". Protocols must match.
x.each.contents @ jquery.min.js?v=66fe24a40a6708181b37ca63996e5991:5x.extend.map @ jquery.min.js?v=66fe24a40a6708181b37ca63996e5991:4x.each.x.fn.(anonymous function) @ jquery.min.js?v=66fe24a40a6708181b37ca63996e5991:5(anonymous function) @ mail.js?v=66fe24a40a6708181b37ca63996e5991:940x.event.dispatch @ jquery.min.js?v=66fe24a40a6708181b37ca63996e5991:5x.event.add.v.handle @ jquery.min.js?v=66fe24a40a6708181b37ca63996e5991:5
```

wurstchristoph:
@LukasReschke as far as I have tested, yes (using Firefox)...

jancborchardt:
@wurstchristoph always test with both FF and Chrome. ;)

LukasReschke:
What I can think about as alternative is to somehow preload the pages and set regular HTTP caching headers on the HTML messages.

PR #778, opened 2015-06-16
resize iframe on window resize

### LINKED ISSUE #573: iframe needs to be resized on parent window resize
When the browser window gets resized, the width of the iframe changes, reflowing the content of the message and causing the height to change. However, that height doesn't get updated, so a second scrollbar can show up inside the iframe. We should run a handler on window resize that updates the height property of the iframe, and probably some CSS to prevent the scrollbar showing up between updates.

cc @jancborchardt

wurstchristoph:
fixes #573 
@Xenopathic @jancborchardt @aidanamavi please review

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -530,6 +530,9 @@ var Mail = {
 				}
 			});
 
+			// Register UI events
+			window.addEventListener('resize', Mail.UI.Events.onWindowResize);
+
 			Marionette.TemplateCache.prototype.compileTemplate = function(rawTemplate) {
 				return Handlebars.compile(rawTemplate);
 			};
@@ -1120,6 +1123,12 @@ var Mail = {
 				Mail.BackGround.messageFetcher.restart();
 				// hide message detail view on mobile
 				$('#mail-message').addClass('hidden-mobile');
+			},
+
+			onWindowResize: function() {
+				// Resize iframe
+				var iframe = $('#mail-content iframe');
+				iframe.height(iframe.contents().find('html').height() + 20);
 			}
 		};
 


js/mail.js
```diff
@@ -1120,6 +1123,12 @@ var Mail = {
 				Mail.BackGround.messageFetcher.restart();
 				// hide message detail view on mobile
 				$('#mail-message').addClass('hidden-mobile');
+			},
+
+			onWindowResize: function() {
+				// Resize iframe
+				var iframe = $('iframe');
```
Xenopathic: I recommend making this more specific, to avoid accidentally catching other iframes: `$('#mail-content iframe')`

wurstchristoph: Makes sense. I updated my PR ;-)

Xenopathic:
Works :+1:

wurstchristoph:
> I can't seem to replicate these issues with the latest masters of Mail and ownCloud on both Firefox and Chrome. When the browser window is resized, the iFrame's height is always correct; if necessary, the iFrame horizontal scroll bar will adjust.

> With this current fix, there is no noticeable change with Firefox; however, with Chrome the javascript will add 20 height pixels every time the browser's window is adjusted, which leaves large amounts of white space at the end of the iFrame content and before the horizontal iFrame scroll bar.
:-1:

@aidanamavi It works for me on FF (Linux and Windows):
# master branch
oc1
oc2

# resize-iframe branch
oc3
oc4

Those 20 extra pixels should be removed (I copied the code from the iframe load function)

wurstchristoph:
@aidanamavi are you sure you tested a HTML message? There are no issues with non-HTML messages

aidanamavi:
Positive. I'm running on Mac OS.

wurstchristoph:
> Positive. I'm running on Mac OS.
Ok, that could be the reason ;-)

aidanamavi:
Could be. The original issue didn't state a specific browser or OS, so I presumed this issue was testable by everyone.

It would be helpful to know the browser and OS of the issue next time, so people don't waste time testing untestable issues. I'll be sure to ask for clarification in the future.

wurstchristoph:
@aidanamavi even if the issue did state a specific browser or OS there is a chance it can be reproduced on the same browser on another OS.

Xenopathic:
Test on any email with text that reflows when the window is resized, be it an HTML email or a plain text email. The height is set on the iframe to match the content, and after this PR when the window is resized that height gets reset to the right value.

jancborchardt:
Supernice – fixes the issue, no double scrollbars anymore on resize! :+1:

jancborchardt:
@wurstchristoph @aidanamavi so do you want to also fix the whitespace issue or should we merge this in?

aidanamavi:
@jancborchardt @wurstchristoph I took a second look at this fix today and everything is working as you would expect. I'm not sure what was going on yesterday, but I apologize for any confusion.
:+1:

jancborchardt:
@aidanamavi no worries - good that it works now! :)

PR #786, opened 2015-06-17
Fix saving attachments to files

### LINKED ISSUE #438: Attachment name wrong on download & save to files (only on one mail server)

DeepDiver1975:
@jancborchardt 

fixes #438


jancborchardt:
Awesome, fixes the issue :+1:

PR #790, opened 2015-06-17
Fix PHPDoc

LukasReschke:
Made little autists crazy when their IDE is always yellow

@jancborchardt Please review. (just merge - that's enough :speak_no_evil:)


jancborchardt:
I only understand »train station«. This seems to be something for @Gomez and @clementhk :D

Gomez:
:+1:

PR #792, opened 2015-06-17
special notification layout if there is only 1 new message, fix #710

### LINKED ISSUE #710: Special notification style when there is only one email
When there is only one new email, we shouldn’t waste the first line to say
> 1 new message in Inbox
> from `Sender`

But rather say (basically same as in the message list in the app):
> `Sender`
> `Subject`

That will make the notifications more useful because then you more quickly see what a notification is about in the very common case of just one new email having rolled in.

@DeepDiver1975

jancborchardt:
First work in progress to fix #710 – not tested yet. :) cc @colmoneill @aidanamavi @DeepDiver1975 if you wanna check it out.

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -228,13 +228,26 @@ var Mail = {
 				} else {
 					from = from.slice(0, 2);
 				}
-				var body = n('mail',
-					'%n new message in {folderName} \nfrom {from}',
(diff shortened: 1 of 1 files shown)


jancborchardt:
:+1: nice!

aidanamavi:
:+1: Awesome work, gentlemen. :)

Issue #795, opened 2015-06-17
Unified composer view


ChristophWurst:
ATM I know of at least 2 different composer views (+ different js view code for each of them), but we will need more of them:
- new message composer
- reply message composer
- forward message composer
- mailto-link message composer

To make things nice and easy we should combine all of them (or reuse the first one), so everywhere we use a composer, the same features are available (e.g. HTML content, drafts, attachment)

this is needed for at least the following issues #95, #602, #603, #605, #734, #735

DeepDiver1975:
Yes - that needs to be done.


I suggest to use the new message composer, add attachments from local computer and finally reuse it then.

PR #800, opened 2015-06-18
Appropriately responds to programmatic input in message bodies.

### LINKED ISSUE #582: Encrypted text inserted by Mailvelope does not resize textarea nor activate send button
There’s a slight issue with Mailvelope we need to look into: The message text pasted into the textarea does not result in the textarea being resized nor the send button being activated:
capture du 2015-05-04 12 28 30

aidanamavi:
Added checks to text area input on message bodies for both new messages and reply messages.
Added responses to programmatic input that resembles client input.

Fixes #582

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -1320,6 +1320,36 @@ $(document).ready(function() {
 	$(document).on('change input paste keyup', '.reply-message-fields #to', Mail.UI.toggleReplyButton);
 	$(document).on('change input paste keyup', '.reply-message-body', Mail.UI.toggleReplyButton);
 
+	/**
+	* Detects pasted text by browser plugins, and other software.
+	* Check for changes in message bodies every second.
+	*/
+	setInterval((function() {
+		// Begin the loop.
+		return function() {
+
+			// Define which elements hold the message body.
+			var MessageBody = $('#new-message-body');
+
+			/**
+			 * If the message body is displayed and has content:
+			 * Prepare the message body content for processing.
+			 * If there is new message body content to process:
+			 * Resize the text area.
+			 * Toggle the send button, based on whether the message is ready or not.
+			 * Prepare the new message body content for future processing.
+			 */
+			if (MessageBody.val()) {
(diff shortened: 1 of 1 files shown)


js/mail.js
```diff
@@ -1306,6 +1306,45 @@ $(document).ready(function() {
 	$(document).on('change input paste keyup', '.reply-message-fields #to', Mail.UI.toggleReplyButton);
 	$(document).on('change input paste keyup', '.reply-message-body', Mail.UI.toggleReplyButton);
 
+	/**
+	* Detects pasted text by browser plugins, and other software.
+	* Check for changes in message bodies every second.
```
jancborchardt: Is this really a good idea performance-wise? Or is this likely to cause lags? It’s not _that_ important if the tradeoff is too big.

aidanamavi: No one will notice any decrease in performance whatsoever.

jancborchardt:
Works very nice except for one issue which was better before: The reply box now always is expanded. Previously it only expanded (to show the full previous message) as soon as you clicked inside of it. That way you could also still see the reply button.

aidanamavi:
Please read the original issue #582. Any input whatsoever should expand the text areas. This is a fix for input detection; which addresses input made by any third party program, not just Mailvelope.

The new message and reply message composer are also intended to function the same #795.

If you always want to see the reply button while composing, open a new issue to discuss and address it. (BTW, I want to see this feature, too, and would love to discuss this with you in another issue.)

jancborchardt:
Right, ok. :+1: then :)

jancborchardt:
Second review please @ChristophWurst :)

ChristophWurst:
Good catch @jancborchardt, I also like that textareas do not expand by default at the moment.

@aidanamavi could you try to make this work again? Otherwise we would break a feature on purpose an then have to add another fix to bring it back...

aidanamavi:
The latest push squashes the previous commits on the branch into one.

ChristophWurst:
Great :+1:

PR #806, opened 2015-06-18
remove deprecated params calls; add some unit tests

ChristophWurst:
``$this->params(..)`` is deprecated, so I removed all the calls and added method parameters

@DeepDiver1975 @LukasReschke @Xenopathic review please :)

### DIFF
--- a/lib/controller/accountscontroller.php
+++ b/lib/controller/accountscontroller.php
@@ -83,12 +83,12 @@ class AccountsController extends Controller {
 	/**
 	 * @param string $appName
 	 * @param \OCP\IRequest $request
-	 * @param $mailAccountMapper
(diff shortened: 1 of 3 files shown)


DeepDiver1975:
We have to test this on oc7.

DeepDiver1975:
> yes - but we do not test appframeworks phpdoc parsing to get an understanding of parameter types - may guts feeling is telling be that this feature was not available in oc7

I love version information on our public api - https://github.com/owncloud/core/blob/master/lib/public/appframework/controller.php#L157

so this change is okay :+1:

LukasReschke:
Ah you refered to that - stupid me :gun:

LukasReschke:
:+1:

PR #811, opened 2015-06-19
[WIP] Attachments from local

DeepDiver1975:
:one: local file can be attached now and it will properly be sent.

This works WITHOUT jquery.upload - but we should try to make it work wit it.

Because if heavy js kung-fu  I'm out of this game ....

pilot-has-nothing-to-do-here_o_1330229


itheiss:
Is this ready for testing or still WIP?

DeepDiver1975:
Still wip. Feel free to take over.

PR #814, opened 2015-06-20
Bool -> boolean

### LINKED ISSUE #813: Mail is always using AutoConfig
The check on line https://github.com/owncloud/mail/blob/master/lib/controller/accountscontroller.php#L199 dosent work. $_POST elements are strings, so this IF can never be "false".

Edit: Tested with the current commit 1695a91dc50d69ca7c08771c290f840104224502 using owncloud  https://github.com/owncloud/core/commit/5ed9743570173fbea59f150c2fe8da8d6c681cba

DeepDiver1975:
fixes #813


ChristophWurst:
replaced by #815

PR #815, opened 2015-06-20
fix PHPDoc parameter types for proper casting to bool/int

### LINKED ISSUE #813: Mail is always using AutoConfig

ChristophWurst:
@DeepDiver1975 @Nawuko please review

fixes #813 and replaces #814 

Additionally to ``$autoDetect``, also integer parameters like ``$imapPort`` are currently not being casted. This should be fixed with this PR

### DIFF
--- a/lib/controller/accountscontroller.php
+++ b/lib/controller/accountscontroller.php
@@ -134,7 +134,7 @@ class AccountsController extends Controller {
 	/**
 	 * @NoAdminRequired
 	 *
-	 * @param $accountId
+	 * @param int $accountId
 	 * @return JSONResponse
 	 */
 	public function show($accountId) {
@@ -159,7 +159,7 @@ class AccountsController extends Controller {
 	/**
 	 * @NoAdminRequired
 	 *
-	 * @param string $accountId
+	 * @param int $accountId
 	 * @return JSONResponse
 	 */
 	public function destroy($accountId) {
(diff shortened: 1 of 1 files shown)


ChristophWurst:
@Raydiation is there a way to unit test this typecasting?

DeepDiver1975:
:+1:

Raydiation:
As for the type cast unit testing question: You'd have to write a utility that does that reflection and casts by yourself if you want to unit test this. 

However I think that this would be better tested in an integration or acceptance test since you are actually testing the interfaces. Additionally PHP 7 will allow you to add scalar type hints that do exactly that, which basically means: the problem will solve itself after some time. I'm planning to add a PR that prefers the scalar types over the annotations once I get my hands on a stable PHP 7 :)

Raydiation:
https://github.com/owncloud/core/issues/17076

ChristophWurst:
@Xenopathic @clementhk @LukasReschke please review :-)

Issue #602, opened 2015-05-05
Store reply drafts on IMAP server


ChristophWurst:
Currently reply message are simply cached in local storage. For best integration with other mail clients, those draft should also be saved to IMAP

jancborchardt:
@wurstchristoph @irgendwie someone of you want to work on this for 0.2? It would indeed be good to finish up the draft stuff.

ChristophWurst:
@jancborchardt I'll try to finish it in 0.2. First step: #819

ChristophWurst:
fixed with PR #819

Issue #568, opened 2015-05-01
Add BCC in reply form


colmoneill:
The reply form at the end of a displayed mail has a carbon copy field, but not a blackened cc field.

jancborchardt:
Right, good call. We must’ve forgot. :D

jancborchardt:
This is also blocked by the rework of the message / reply fields regarding adding attachments from the computer: https://github.com/owncloud/mail/issues/462 – moving to 0.3 also because it’s less important than the other stuff.

ChristophWurst:
fixed with PR #819

Issue #460, opened 2015-04-10
Ability to add attachments when replying


jancborchardt:
Currently adding attachments is only possible for new messages. Should also be possible for replies.

Xenopathic:
This seems like a bug - leaving on 0.1

jancborchardt:
Yeah, it’s definitely expected to work as it does also for new mails.

jancborchardt:
This would really be good for 0.1 @DeepDiver1975

jancborchardt:
Can we do this for 0.1 then? Probably doesn't involve as much jQuery as #462

DeepDiver1975:
> Can we do this for 0.1 then? Probably doesn't involve as much jQuery as #462

doable

Xenopathic:
I've tried to get started on this, but the code is an absolute mess! There's sending logic in js/send-mail.js for replies and in js/views/sendmail.js for new mails, templates/compose.php seems completely broken at the moment, and there are so many hooks and events all over the place it's getting hard to keep track of things.

Xenopathic:
So here's the idea - we have a Handlebars template for new messages, which I will extend so we can pre-fill fields (like the To and Subject), then use that same template not just for new messages but underneath existing mails for replies. Also means the user has more control over email replies (manual Subject editing for example).

jancborchardt:
I would still prefer to by default hide some fields in the case of reply, such as the subject. But we’ll see about that when we’re at it. I think it’s a great idea to unify it as stuff like #568 (add bcc for replies) will also be fixed. :)

ChristophWurst:
fixed with PR #819

PR #819, opened 2015-06-22
make composer view reusable, reply IMAP drafts, mailto IMAP drafts, reply attachment

### LINKED ISSUE #795: Unified composer view

ChristophWurst:
~~first step towards #795~~ this is #795 :rocket: 

~~My intention is to neither change nor break any functionality with this PR, I only want to make the new-message-composer view ready for using it as reply composer too (and then implement #602)~~
As this is much harder than I thought in the first place, I'll focus on not breaking the _new message composer_ functionality, other composers (e.g. mailto, reply) will be removed/replaced anyway.

Things to do:
- [x] init
- [x] reset
- [x] send mail
- [x] save IMAP draft
- [x] HTML classes/IDs
- [x] use backbone [$-selector](http://backbonejs.org/#View-dollar)
- [x] fix autocompletion
- [x] add draft UID when sending the message
- [x] hide subject when opening replies
- [x] send replies
- [x] create IMAP reply drafts
- [x] fix reply composer message body size
- [x] expand reply body on click
- [x] reply to/cc not filled correctly
- [x] submit button name is 'Send' for both new messages and replies
- [x] reply cc/bcc toggle slides down and up again (after opening cc/bcc in the new-message-composer)
- [x] fix attachment
- [x] test and fix mailto composer
- [x] add missing reply subject
- [ ] mailto composer does not show attachment names, ``{{displayName}}``is shown instead

fixes #795, #460, #602, #568


jancborchardt:
Wohoo, major improvement. :rocket: :)

ChristophWurst:
At last I finished this PR. This was more work to do than I expected, but no we have one composer for new messages, replies and the mailto compose window. All of them finally support drafts and attachments :smile: 

There's one known bug: the mailto composer does not show the filename of attached fiels, instead ``{{diplayName}}`` is shown. Could be a Backbone related issue, but I couldn't figure out how to fix it :-/

jancborchardt:
Another bug: When deleting a message, it doesn’t switch to the next newest message. :) Works in master.

ChristophWurst:
> Another bug: When deleting a message, it doesn’t switch to the next newest message. :) Works in master.

@jancborchardt yes, there's a bug. However, I could also reproduce this error on my production system (master branch):
bildschirmfoto von 2015-06-29 17-09-09

jancborchardt:
Hm, works fine for me in master at least.

ChristophWurst:
@jancborchardt should be fixed now ;-)

The TypeError still exists, but seems not to be related to this PR. Is thrown in FF but not in Chrome btw ;-)

js/views/composer.js
```diff
@@ -0,0 +1,369 @@
...
+	folderId: null,
+	messageId: null,
+
+	draftInterval: 1500,
+	draftTimer: null,
+	draftUID: null,
+	hasData: false,
+	autosized: false,
+
+	events: {
```
DeepDiver1975: Really nice! I wanted to get rid of all those jquery-fu in mail.js for long time! 

@ChristophWurst 

youre-awesome

DeepDiver1975:
@ChristophWurst travis is crying because of this

````
js/send-mail.js: line 1, col 1, 'Mail' is defined but never used.

DeepDiver1975:
I did not yet test this - but the code looks really great! Awesome job!

ChristophWurst:
> @ChristophWurst travis is crying because of this
> 
> js/send-mail.js: line 1, col 1, 'Mail' is defined but never used.

I removed it, now travis should be happy again :-)

DeepDiver1975:
let's fix the mailto composer in a second pr - this is awesome - time to move it in :+1:

PR #824, opened 2015-06-24
Display inline images - fixes #138

### LINKED ISSUE #138: Properly handle html  mails with embedded attachments
as of today we show embedded images of an html mail as attachment - but it should be displayed within the html mail

summoning @LukasReschke regarding security impacts

DeepDiver1975:
Fix #138

### DIFF
--- a/lib/message.php
+++ b/lib/message.php
@@ -22,6 +22,7 @@
 
 namespace OCA\Mail;
 
+use Closure;
 use Horde_Imap_Client;
 use Horde_Imap_Client_Data_Fetch;
 use OCA\Mail\Service\Html;
@@ -424,9 +425,13 @@ class Message {
 	}
 
 	/**
+	 * @param int     $accountId
+	 * @param string  $folderId
+	 * @param int     $messageId
+	 * @param Closure $attachments
 	 * @return string
 	 */
-	public function getHtmlBody($accountId, $folderId, $messageId, $attachments) {
+	public function getHtmlBody($accountId, $folderId, $messageId, Closure $attachments) {
 		return $this->htmlService->sanitizeHtmlMailBody($this->htmlMessage, [
 			'accountId' => $accountId,
 			'folderId' => $folderId,
--- a/lib/message.php
+++ b/lib/message.php
@@ -332,7 +332,8 @@ class Message {
 				'messageId' => $this->messageId,
 				'fileName' => $filename,
 				'mime' => $p->getType(),
-				'size' => $p->getBytes()
+				'size' => $p->getBytes(),
+				'cid' => $p->getContentId()
 			];
 			return;
 		}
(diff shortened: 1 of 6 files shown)


jancborchardt:
Works as they are displayed inline, but they are also still shown as attachments. We don’t want them to show as attachments, right?

DeepDiver1975:
> We don’t want them to show as attachments, right?

well - this allows to download and save them to the files app.
We loose functionality if we no longer show them as attachments

jancborchardt:
Right. Mails with those types of images also tend to be photos and such (as opposed to logos and the usual HTML mail stuff)

 :+1: in that case.

jancborchardt:
Another review please @LukasReschke @aidanamavi @ChristophWurst @irgendwie :)

lib/service/htmlpurify/cidurischeme.php
```diff
@@ -0,0 +1,41 @@
...
+
+	/**
+	 * Validates the components of a URI for a specific scheme.
+	 *
+	 * @param HTMLPurifier_URI $uri Reference to a HTMLPurifier_URI object
+	 * @param HTMLPurifier_Config $config
+	 * @param HTMLPurifier_Context $context
+	 * @return bool success or failure
+	 */
+	public function doValidate(&$uri, $config, $context) {
```
ChristophWurst: Anything keeping us from using [PHP type hints](https://secure.php.net/manual/en/language.oop5.typehinting.php) here?

lib/service/htmlpurify/transformurlscheme.php
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	/**
+	 * Transformator which will rewrite all HTTPS and HTTP urls to
+	 * @param \HTMLPurifier_URI $uri
+	 * @param HTMLPurifier_Config $config
+	 * @param \HTMLPurifier_Context $context
+	 * @return bool
+	 */
+	public function filter(&$uri, $config, $context) {
```
ChristophWurst: same here

lib/service/htmlpurify/transformurlscheme.php
```diff
@@ -0,0 +1,105 @@
...
+
+		return true;
+	}
+
+	/**
+	 * @param $uri
+	 * @param $context
+	 * @return HTMLPurifier_URI
+	 */
+	private function filterHttp(&$uri, $context) {
```
ChristophWurst: and here

lib/message.php
```diff
@@ -425,8 +426,12 @@ public function getListArray() {
 	/**
 	 * @return string
 	 */
-	public function getHtmlBody() {
-		return $this->htmlService->sanitizeHtmlMailBody($this->htmlMessage);
+	public function getHtmlBody($accountId, $folderId, $messageId, $attachments) {
```
ChristophWurst: ``$attachments`` is a closure, isn't it? We should type-hint it and add add info to the doc block. It's not clear how to use this method otherwise

ChristophWurst:
Everything else looks good :+1:

lib/service/htmlpurify/cidurischeme.php
DeepDiver1975: these methods overwrite methods of the base class - so we better no mess with them

lib/message.php
DeepDiver1975: indeed - thx @ChristophWurst feel free to fix and push thx

PR #830, opened 2015-06-29
Unread/read status syncing, message list syncing, and infinite scrolling w/feedback.

aidanamavi:
Added event handler for message click to toggle the unseen flag.
Added syncing for the message unseen flag to the mail server.
Added unseen status fetcher for the Inbox folder.
Added message list syncing.
Added infinite scrolling message list with feedback.

Fixes #773	Detect flag changes and expunged messages
Fixes #499	Automatic lazy loading of more messages when scrolled to the end of the list
Fixes #104	Pull to refresh & automatic update of messages
Related #633	"Check messages" should sync deleted messages as well.

PR #633, opened 2015-05-06
use handlebars to change css class of the currently opened folder

wurstchristoph:
fixes #461 
@jancborchardt

PR #834, opened 2015-06-30
Unread/read status syncing, message syncing, and infinite scrolling.

### LINKED ISSUE #499: Automatic lazy loading of more messages when scrolled to the end of the list
Currently when you scroll down the message list you need to manually click »Load more …«. Instead, more mails should automatically get loaded shortly before you reach the end of the list.

Similar to pull to refresh / automatically loading newer messages: https://github.com/owncloud/mail/issues/104

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/11391226-automatic-lazy-loading-of-more-messages-when-scrolled-to-the-end-of-the-list?utm_campaign=plugin&utm_content=tracker%2F216487&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F216487&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #773: [master] under yet unknown conditions the check-messages-button disappears

### LINKED ISSUE #803: Remaining issues for unified inbox
Sorted by importance

BUGS
* [x] On new mails, the list is not automatically refreshed

ENHANCEMENTS
* [x] We need to not show the color of the account when there is only one. Simply do not set the background-color of `.mail-account-color`
* [x] The title says »All inboxes - - Mail - ownCloud« (note the »- -« double dash, should be just a single one because there’s no account to show)

Ref https://github.com/owncloud/mail/pull/750

aidanamavi:
Added event handler for message click to toggle the unseen flag.
Added syncing for the message unseen flag to the mail server.
Added unseen status fetcher for the Inbox folder.
Added message list syncing.
Added infinite scrolling message list with feedback.

Fixes #104	Pull to refresh & automatic update of messages
Fixes #499	Automatic lazy loading of more messages when scrolled to the end of the list
Fixes #633	"Check messages" should sync deleted messages as well.
Fixes #773	Detect flag changes and expunged messages
Fixes #830	Adds bug fixes addressed in this pull request
Fixes #803	On new mails, the list is not automatically refreshed
			The 'Check messages' button disappears once refreshing the message list has finished.
			When clicking »load more« the list jump-scrolls to the top as soon as more emails are loaded

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -32,6 +32,14 @@ $app->registerRoutes($this,
 				'url' => '/accounts/{accountId}/folders/{folderId}/messages/{messageId}/toggleStar',
 				'verb' => 'POST'],
 			[
+				'name' => 'messages#toggleUnseen',
+				'url' => '/accounts/{accountId}/folders/{folderId}/messages/{messageId}/toggleUnseen',
+				'verb' => 'POST'],
+			[
+				'name' => 'messages#getUnseenCount',
+				'url' => '/accounts/{accountId}/folders/{folderId}/getUnseenCount',
+				'verb' => 'POST'],
+			[
 				'name' => 'proxy#redirect',
 				'url' => '/redirect',
 				'verb' => 'GET'],
--- a/css/mail.css
+++ b/css/mail.css
@@ -224,20 +224,16 @@
 	overflow-x: hidden;
 	overflow-y: auto;
 }
-#load-new-mail-messages,
 #load-more-mail-messages {
 	display: none;
-	margin: 10px auto;
-	padding: 10px 10px 6px;
-}
-#load-more-mail-messages {
-	margin-bottom: 250px;
-}
-/* TODO: put this in core icons.css as general rule for buttons with icons */
-#load-new-mail-messages.icon-loading-small,
-#load-more-mail-messages.icon-loading-small {
-	padding-left: 32px;
-	background-position: 9px center;
+	height: 68px;
+	border-top: 1px solid #eee;
+	-webkit-box-shadow: inset 0 4px 9px -6px rgba(0,0,0,.8);
+	-moz-box-shadow: inset 0 4px 9px -6px rgba(0,0,0,.8);
+	box-shadow: inset 0 4px 9px -6px rgba(0,0,0,.8);
+	text-align: center;
+	line-height: 68px;
+	font-size: 16px;
 }
 
 .mail-message-header {
--- a/css/mail.css
+++ b/css/mail.css
@@ -228,9 +228,6 @@
 	display: none;
 	height: 68px;
 	border-top: 1px solid #eee;
-	-webkit-box-shadow: inset 0 4px 9px -6px rgba(0,0,0,.8);
-	-moz-box-shadow: inset 0 4px 9px -6px rgba(0,0,0,.8);
-	box-shadow: inset 0 4px 9px -6px rgba(0,0,0,.8);
 	text-align: center;
 	line-height: 68px;
 	font-size: 16px;
--- a/js/mail.js
+++ b/js/mail.js
@@ -585,9 +585,6 @@ var Mail = {
 			});
 			Mail.State.folderView.render();
 
-			Mail.State.folderView.listenTo(Mail.UI.messageView, 'change:unseen',
-				Mail.State.folderView.changeUnseen);
-
 			// request permissions
 			if (typeof Notification !== 'undefined') {
 				Notification.requestPermission();
@@ -597,6 +594,11 @@ var Mail = {
 				OC.Plugins.register('OCA.Search', Mail.Search);
 			}
 
+			/**
+			 * Checks and loads any new messages and changes to the current folder.
+			 * @todo combine this with the checkForNotifications method.
+			 * @todo leadMessages needs to update the unseen folder counter.
+			 */
 			setInterval(Mail.BackGround.checkForNotifications, 5*60*1000);
 			this.loadAccounts();
 		};
@@ -700,8 +702,6 @@ var Mail = {
 			$('#folders').removeClass('hidden');
 			$('#mail-setup').addClass('hidden');
 
-			$('#load-new-mail-messages').hide();
-			$('#load-more-mail-messages').hide();
 			$('#emptycontent').hide();
 
 			if (noSelect) {
@@ -729,29 +729,22 @@ var Mail = {
 							if (jsondata.length > 0) {
 								Mail.UI.addMessages(jsondata);
 
+								// Load once to initiate object accessability outside of Backbone.
+								Mail.UI.messageView.loadMessages('newFolder');
+
 								// Fetch first 10 messages in background
 								var first10 = Mail.State.messageView.collection.first(10);
 								_.each(first10, function (message) {
 									Mail.BackGround.messageFetcher.push(message.id);
 								});
 
+								// Load the first message in the message list.
 								var messageId = Mail.State.messageView.collection.first().get('id');
 								Mail.UI.loadMessage(messageId);
-								// Show 'Load More' button if there are
-								// more messages than the pagination limit
-								if (jsondata.length > 20) {
-									$('#load-more-mail-messages')
-										.fadeIn()
-										.css('display', 'block');
-								}
 							} else {
 								$('#emptycontent').show();
 								$('#mail-message').removeClass('icon-loading');
 							}
-							$('#load-new-mail-messages')
-								.fadeIn()
-								.css('display', 'block')
-								.prop('disabled', false);
 
 						},
 						error: function(textStatus) {
@@ -963,8 +956,6 @@ var Mail = {
 				mailBody
 					.html(html)
 					.removeClass('icon-loading');
-
-				Mail.UI.messageView.setMessageFlag(messageId, 'unseen', false);
 
 				// HTML mail rendering
 				$('iframe').load(function() {
@@ -1195,6 +1186,35 @@ var Mail = {
 
 $(document).ready(function() {
 	Mail.UI.initializeInterface();
+
+	// Enables infinite scroll ability.
+	setInterval(function() {
+
+		// Element identifies the element we want to trigger our loading on.
+		//var element = $('.mail_message_summary:nth-last-child(1)');
+		var element = $('#load-more-mail-messages');
+		var outerHeight = element.outerHeight();
+		var offset = element.offset();
+		var scrollLength = $('#mail_messages').scrollTop();
+		var screenHeight = $(document).height();
+		var messageListHeight = scrollLength + offset.top + outerHeight;
+
+		if ((scrollLength + screenHeight) >= messageListHeight) {
+			Mail.UI.messageView.loadMessages('addMore');
+		}
+
+	}, 1000);
+
+	/**
+	 * Checks and loads any new messages and changes to the current folder.
+	 * @todo combine this with the checkForNotifications method.
+	 */
+	setInterval((function() {
+		// Begin the loop.
+		return function() {
+			Mail.UI.messageView.loadMessages();
+		};
+	})(), 10*1000);
 
 	// auto detect button handling
 	$('#auto_detect_account').click(function(event) {
--- a/js/send-mail.js
+++ b/js/send-mail.js
@@ -122,7 +122,8 @@ $(function() {
 			},
 			type: 'POST',
 			success:function() {
-				Mail.UI.messageView.setMessageFlag(messageId, 'answered', true);
+				// Reload the message list to display the answered flag.
+				Mail.UI.messageView.loadMessages();
 				OC.Notification.showTemporary(t('mail', 'Message sent!'));
 
 				// close reply
--- a/js/views/folder.js
+++ b/js/views/folder.js
@@ -97,14 +97,41 @@ views.Folders = Backbone.Marionette.CollectionView.extend({
 		return activeFolder;
 	},
 
-	changeUnseen: function(model, unseen) {
+	/**
+	 * Retrieve and/or adjust the unseen count for the current folder on display.
+	 * @todo Add the ability to adjust any folder.
+	 * @param object $model The model to reference for the active folder.
+	 * @param boolean|string $unseen True/false for manual unseen entries; check for updates retrieved from backend.
+	 */
+	changeUnseenCount: function(model, unseen) {
 		// TODO: currentFolderId and currentAccountId should be an attribute of this view
+		unseen = unseen || 'check';
 		var activeFolder = this.getFolderById();
-		if (unseen) {
-			activeFolder.set('unseen', activeFolder.get('unseen') + 1);
-		} else {
-			if (activeFolder.get('unseen') > 0) {
-				activeFolder.set('unseen', activeFolder.get('unseen') - 1);
+		// Ask the IMAP server what the unseen count is.
+		if (unseen === 'check') {
+			$.ajax(
+				OC.generateUrl('apps/mail/accounts/{accountId}/folders/{folderId}/getUnseenCount',
+				{ accountId: Mail.State.currentAccountId,
+					folderId: Mail.State.currentFolderId,
+				}), {
+					type:'POST',
+					success: function (jsondata) {
+						// Set the new unseen count.
+						activeFolder.set('unseen', jsondata.unseen);
+					},
+					error: function() {
+						Mail.UI.showError(t('mail', 'Could not get unread count for this folder. Please try again.'));
+					}
+				});
+		}
+		// Adjust the display of the unseen count manually.
+		if (typeof unseen === 'boolean') {
+			if (unseen) {
+				activeFolder.set('unseen', activeFolder.get('unseen') + 1);
+			} else {
+				if (activeFolder.get('unseen') > 0) {
+					activeFolder.set('unseen', activeFolder.get('unseen') - 1);
+				}
 			}
 		}
 		this.updateTitle();
--- a/js/views/message.js
+++ b/js/views/message.js
@@ -19,9 +19,15 @@ views.Message = Backbone.Marionette.ItemView.extend({
 	},
 
 	events: {
-		"click .action.delete" : "deleteMessage",
 		"click .mail-message-header" : "openMessage",
+		"click .action.delete" : "deleteMessage",
 		"click .star" : "toggleMessageStar"
+	},
+
+	initialize: function (){
+		// AFAIK, this changes nothing.
+		// @todo look into the purpose, and need of this bind.
+		this.model.bind('change', this.render, this);
 	},
 
 	onRender: function () {
@@ -41,13 +47,14 @@ views.Message = Backbone.Marionette.ItemView.extend({
 	},
 
 	toggleMessageStar: function(event) {
+		// Stop any event handlers below this element from being fired.
 		event.stopPropagation();
 
-		var messageId = this.model.id;
-		var starred = this.model.get('flags').get('flagged');
 		var thisModel = this.model;
+		var messageId = this.model.id;
+		var starred = this.model.attributes.flags.flagged;
 
-		// directly change star state in the interface for quick feedback
+		// Change star state in the interface for quick feedback.
 		if(starred) {
 			this.ui.star
 				.removeClass('icon-starred')
@@ -70,10 +77,12 @@ views.Message = Backbone.Marionette.ItemView.extend({
 				},
 				type:'POST',
 				success: function () {
-					thisModel.get('flags').set('flagged', !starred);
+					// Retrieve the starred/flag status from the backend and display the results.
+					Mail.UI.messageView.loadMessages();
 				},
 				error: function() {
 					Mail.UI.showError(t('mail', 'Message could not be starred. Please try again.'));
+					// Change star state in the interface to it's previous state.
 					thisModel.get('flags').set('flagged', starred);
 				}
 			});
@@ -85,6 +94,66 @@ views.Message = Backbone.Marionette.ItemView.extend({
 		Mail.UI.loadMessage(this.model.id, {
 			force: true
 		});
+
+		var message = this.model;
+		this.setMessageFlag(message, 'unseen');
+	},
+
+	/**
+	 *
+	 * @todo Combine toggleMessageStar.
+	 * @todo Fully integrate object handlers.
+	 * @todo add flag handlers for: flagged, answered, deleted, draft, forwarded, hasAttachments
+	 * @param object $message Uses the object for obtaining the ID.
+	 * @param string $flag unseen
+	 * @param boolean $value Default is to toggle, otherwise use defined value.
+	 */
+	setMessageFlag: function(message, flag, value) {
+		var messageId = message.id;
+		var unseen;
+		message = this.$el;
+
+		// Reusable function that provides immediate feedback while AJAX is loading.
+		var setUnseenCss = function (unseen) {
+			if (unseen) {
+				message.removeClass('unseen');
+			} else {
+				message.addClass('unseen');
+			}
+		};
+
+		if (flag === 'unseen') {
+			// If there is no manual selection/value, then toggle.
+			if (_.isUndefined(value)) {
+				unseen = !message.is('.unseen');
+				setUnseenCss(unseen);
+			} else {
+				unseen = value;
+				setUnseenCss(unseen);
+			}
+			$.ajax(
+				OC.generateUrl('apps/mail/accounts/{accountId}/folders/{folderId}/messages/{messageId}/toggleUnseen',
+				{	accountId: Mail.State.currentAccountId,
+					folderId: Mail.State.currentFolderId,
+					messageId: messageId
+				}), {
+					data: {
+						unseen: unseen
+					},
+					type:'POST',
+					success: function () {
+						// Refresh the message list to display the updates.
+						Mail.UI.messageView.loadMessages();
+					},
+					error: function() {
+						Mail.UI.showError(t('mail', 'Could not communicate with the mail server. Please try again.'));
+						// Change unseen state in the interface to it's previous state.
+						setUnseenCss(unseen);
+					}
+				});
+		} else {
+			// @todo Add new flag methods here.
+		}
 	},
 
 	deleteMessage: function(event) {
@@ -92,8 +161,6 @@ views.Message = Backbone.Marionette.ItemView.extend({
 		var thisModel = this.model;
 		this.ui.iconDelete.removeClass('icon-delete').addClass('icon-loading');
 		$('.tipsy').remove();
-
-		thisModel.get('flags').set('unseen', false);
 
 		this.$el.addClass('transparency').slideUp(function() {
 			$('.tipsy').remove();
@@ -140,8 +207,6 @@ views.Message = Backbone.Marionette.ItemView.extend({
 				}
 			});
 	}
-
-
 });
 
 views.NoSearchResultMessageListView = Marionette.ItemView.extend({
@@ -152,7 +217,6 @@ views.NoSearchResultMessageListView = Marionette.ItemView.extend({
 	template: "#no-search-results-message-list-template",
 
 	onRender: function() {
-		$('#load-more-mail-messages').hide();
 	}
 });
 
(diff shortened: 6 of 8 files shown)


aidanamavi:
Reviews please :) @ChristophWurst @DeepDiver1975 @irgendwie @jancborchardt @clementhk

css/mail.css
```diff
@@ -224,20 +224,16 @@
...
-/* TODO: put this in core icons.css as general rule for buttons with icons */
-#load-new-mail-messages.icon-loading-small,
-#load-more-mail-messages.icon-loading-small {
-	padding-left: 32px;
-	background-position: 9px center;
+	height: 68px;
+	border-top: 1px solid #eee;
+	-webkit-box-shadow: inset 0 4px 9px -6px rgba(0,0,0,.8);
+  -moz-box-shadow: inset 0 4px 9px -6px rgba(0,0,0,.8);
+  box-shadow: inset 0 4px 9px -6px rgba(0,0,0,.8);
```
ChristophWurst: please use tabs for indentation :)

js/mail.js
```diff
@@ -595,8 +592,16 @@ var Mail = {
...
+			/**
+			 * Checks and loads any new messages and changes to the current folder.
+			 * @todo combine this with the checkForNotifications method.
+			 * @todo leadMessages needs to update the unseen folder counter.
+			 */
 			setInterval(Mail.BackGround.checkForNotifications, 5*60*1000);
 			this.loadAccounts();
+
+
+
```
ChristophWurst: remove those three blank lines please

js/mail.js
```diff
@@ -1194,6 +1188,35 @@ var Mail = {
...
+	/**
+	 * Checks and loads any new messages and changes to the current folder.
+	 * @todo combine this with the checkForNotifications method.
+	 */
+	setInterval((function() {
+		// Begin the loop.
+		return function() {
+			Mail.UI.messageView.loadMessages();
+		};
+	})(), 10000*1000);
```
ChristophWurst: There's no need to encapsulate the function,
```
setInterval(function() {
    Mail.UI.messageView.loadMessages();
}, 10000*1000);
```
should do the same and is easier to read ;-)

css/mail.css
aidanamavi: Ah, that pesky atom bug keeps popping up. I'll fix it.

js/mail.js
aidanamavi: Good catch. And I need to reset that refresh rate back down to 10 seconds instead of 10000. ;) I was running error tests and needed to stop the interval, so I temporarily changed the timer.

aidanamavi: The encapsulate function ensures that we don't get the pesky TypeError for an undefined variable in folder.js:75:6. What happens is the interval executes before all scripts are loaded and the encapsulate remedies that issue.

aidanamavi:
@ChristophWurst All fixed, and squashed. :)

ChristophWurst:
I don't think this is intended:
bildschirmfoto von 2015-07-01 11-48-02
(``to`` increasing with each request)

ChristophWurst:
I don't know if this is related to the bug above, but we shouldn't check for changes that often. I guess we can increase the interval to something like 20 to 30 seconds

js/mail.js
```diff
@@ -1194,6 +1185,35 @@ var Mail = {
...
+	/**
+	 * Checks and loads any new messages and changes to the current folder.
+	 * @todo combine this with the checkForNotifications method.
+	 */
+	setInterval((function() {
+		// Begin the loop.
+		return function() {
+			Mail.UI.messageView.loadMessages();
+		};
+	})(), 10*1000);
```
ChristophWurst: this interval should be increased

js/mail.js
ChristophWurst: Are you totally sure about that? As far as I understand what's happening with the self-invoking function, it's executed immediately, which means the inner anonymous function is returned. To me this has the same effect as the my snipped posted above.
Why is there a TypeError and how could the encapsulation help to prevent it? I don't get it :-/

js/mail.js
aidanamavi: Unless you can provide technical feedback, I suggest we leave the interval as it is. Mac Mail and Thunderbird have intervals around this time. I prefer 3 seconds, but I imagine some rare cases where IMAP servers are slow. Tests by @jancborchardt and I revealed that our response time was a fraction of a second. Lets merge this and get some real world results. If anyone reports issues, we can immediately address it then with logic that refreshes based on a finished request as opposed to our time based logic now. (I'd like this function as well, but Jan and I also agree this pull is already quite big.)

aidanamavi:
> (``to`` increasing with each request)
You're reviewing an old branch and commit. That was a WIP issue that I remedied several commits back.

I just rebased this commit, too, so once checks are cleared, it will be good to go.

js/mail.js
ChristophWurst: You still have to remember that client software like Mac mail or Thunderbird are connected to the IMAP server directly, therefore they can simply have an open TCP connection with very little overhead when querying the server. With the mail app we first need to contact the server which then connects to the IMAP server. This means we have to establish two new connections for every request we do, hence it's much slower.

aidanamavi: Comparing the technical aspects has nothing to do with the UX. Our two requests are fast enough to complete in less than about 1/20th the time the interval is set. This is based off current test results. If we have others test results that show this time is not long enough, then we can say it is slow and the interval needs to widen. As it stands, with our tests, it is very fast and 10 seconds is adequate.

ChristophWurst:
I guess this shouldn't happen either:
bildschirmfoto von 2015-07-01 19-56-08

How to reproduce:
1. open the mail app
2. wait for everything to finish loading
3. scroll to the bottom of the message list (inside unified inbox)
4. see how the number of requests grows

My guess: the infinite scroll code loads more messages, even if there has already been sent another one. @aidanamavi could that be the reason?

js/mail.js
```diff
@@ -1196,6 +1187,35 @@ var Mail = {
...
+		var offset = element.offset();
+		var scrollLength = $('#mail_messages').scrollTop();
+		var screenHeight = $(document).height();
+		var messageListHeight = scrollLength + offset.top + outerHeight;
+
+		if ((scrollLength + screenHeight) >= messageListHeight) {
+			Mail.UI.messageView.loadMessages('addMore');
+		}
+
+	}, 1000);
```
ChristophWurst: Instead of polling, could we use [jQuery scroll event](https://api.jquery.com/scroll/) to detect when the user reaches the bottom? This would ensure that there is only one request for loading more messages, as the event shouldn't be triggered again (except the user scrolls up and down again of course)

js/views/folder.js
```diff
@@ -97,14 +97,41 @@ views.Folders = Backbone.Marionette.CollectionView.extend({
...
+						// Set the new unseen count.
+						activeFolder.set('unseen', jsondata.unseen);
+					},
+					error: function() {
+						Mail.UI.showError(t('mail', 'Could not get unread count for this folder. Please try again.'));
+					}
+				});
+		}
+		// Adjust the display of the unseen count manually.
+		if (typeof unseen === 'boolean') {
```
ChristophWurst: please use ``_.isBoolean(unseen)`` (http://underscorejs.org/#isBoolean)

js/views/message.js
```diff
@@ -70,10 +77,12 @@ views.Message = Backbone.Marionette.ItemView.extend({
 				},
 				type:'POST',
 				success: function () {
-					thisModel.get('flags').set('flagged', !starred);
+					// Retrieve the starred/flag status from the backend and display the results.
+					Mail.UI.messageView.loadMessages();
```
ChristophWurst: why do you trigger a reload here?

js/views/message.js
```diff
@@ -152,7 +217,6 @@ views.NoSearchResultMessageListView = Marionette.ItemView.extend({
 	template: "#no-search-results-message-list-template",
 
 	onRender: function() {
-		$('#load-more-mail-messages').hide();
 	}
```
ChristophWurst: No need to keep this empty function, you can remove it completely

js/views/message.js
```diff
@@ -167,8 +231,6 @@ views.Messages = Backbone.Marionette.CompositeView.extend({
 	currentMessageId: null,
 
 	events: {
-		"click #load-new-mail-messages" : "loadNew",
-		"click #load-more-mail-messages" : "loadMore"
 	},
```
ChristophWurst: same here, remove the ``events`` property if it's not used anymore please

ChristophWurst:
Regarding the UI, it's not clear to me why an arrow is shown before the loading spinner is shown. Is this needed? I guess we could also just display some blank space and as soon as new messages are loaded, the spinner is shown.
bildschirmfoto von 2015-07-01 20-35-54
bildschirmfoto von 2015-07-01 20-35-50
Moreover those shadows are inconsistent to any existing styling. While it may look good here, as far as I know, it's not really used anywhere else. @jancborchardt what do you think?

appinfo/routes.php
```diff
@@ -32,6 +32,14 @@
 				'url' => '/accounts/{accountId}/folders/{folderId}/messages/{messageId}/toggleStar',
 				'verb' => 'POST'],
 			[
+				'name' => 'messages#toggleUnseen',
+				'url' => '/accounts/{accountId}/folders/{folderId}/messages/{messageId}/toggleUnseen',
+				'verb' => 'POST'],
+			[
+				'name' => 'messages#getUnseenCount',
+				'url' => '/accounts/{accountId}/folders/{folderId}/getUnseenCount',
+				'verb' => 'POST'],
```
DeepDiver1975: verb POST on a route named get* - please not :wink:

js/mail.js
DeepDiver1975: Yes - I have to agree - we shall not implement our own scroll kung-fu.

My guts feeling is telling me that this PR is already trying to many things.

Please split this up into multiple PRs - THX

js/mail.js
DeepDiver1975: why there yet another call? This is not necessary - the notificatoin check will update the message list properly.

js/send-mail.js
```diff
@@ -122,7 +122,8 @@ $(function() {
 			},
 			type: 'POST',
 			success:function() {
-				Mail.UI.messageView.setMessageFlag(messageId, 'answered', true);
+				// Reload the message list to display the answered flag.
+				Mail.UI.messageView.loadMessages();
```
DeepDiver1975: not necessary to call the server again - we know the message was flagged: so please simply update the flag - as done before

js/views/folder.js
```diff
@@ -97,14 +97,41 @@ views.Folders = Backbone.Marionette.CollectionView.extend({
...
+		unseen = unseen || 'check';
 		var activeFolder = this.getFolderById();
-		if (unseen) {
-			activeFolder.set('unseen', activeFolder.get('unseen') + 1);
-		} else {
-			if (activeFolder.get('unseen') > 0) {
-				activeFolder.set('unseen', activeFolder.get('unseen') - 1);
+		// Ask the IMAP server what the unseen count is.
+		if (unseen === 'check') {
+			$.ajax(
```
DeepDiver1975: why do we call the server here ? we have all the necessary information in the client to compute the count properly.

js/views/folder.js
```diff
@@ -97,14 +97,41 @@ views.Folders = Backbone.Marionette.CollectionView.extend({
...
+				OC.generateUrl('apps/mail/accounts/{accountId}/folders/{folderId}/getUnseenCount',
+				{ accountId: Mail.State.currentAccountId,
+					folderId: Mail.State.currentFolderId,
+				}), {
+					type:'POST',
```
DeepDiver1975: this is no POST but a GET - if needed at all

js/views/message.js
DeepDiver1975: same as above - no need to waste time on the network and create load on the server - we have all required information on the client

js/mail.js
aidanamavi: Every time a user scrolls a pixel, the event is fired. That would be highly inefficient. I tried and tested this method prior to this, and it's a dramatic improvement.

js/views/folder.js
aidanamavi: I'm aware of the Underscore methods, but why use more frameworks if they're unnecessary? Furthermore, you're really holding up this process unnecessarily. You could make these changes yourself, as this is only cosmetic and actually, though unnoticeable, decreases performance.

ChristophWurst: You're right, it's not necessary. On the other hand we use things like ``_.isUndefined(xx)`` although ``xx === undefined`` would do the same (you even used it in your PR). The point is that we should try to be consistent with using helper functions or not, I don't think we should mix it up.

js/mail.js
DeepDiver1975: hmmm - there should be ways to configure the scrolling behavior - we should also have a look at existing js libs which can be customized.

clementhk: Use the combination of both timeout and on scroll.
See http://stackoverflow.com/questions/9144560/jquery-scroll-detect-when-user-stops-scrolling .

js/mail.js
aidanamavi: @DeepDiver1975 If you read the code, and/or description of the pull request, you'd realize this is for syncing. The notification check does not sync the message list; only new messages that arrived after loading the app.

js/views/folder.js
jancborchardt: (Just as a sidenote: We rarely directly change things in the review process of other people's code. Because one important part of reviewing is learning and doing it. Also to not just mess with other folks' code. I know sometimes this can be frustrating but it will make future reviews easier.)

js/mail.js
aidanamavi: @DeepDiver1975 You would be willing to implement hundreds of lines of code from a new library over several lines of code. Then you call this elegant solution kung-fu in comparison?

@clementhk How is firing hundreds to thousands of events a second better than 1 event per second? @Xenopathic has mentioned using this app on a Blackberry, and that is was lagging on there. We should strive for efficiency, should we not?

js/views/folder.js
aidanamavi: That's all fine and well. Before we start judging each others code on coding style, we all as collaborators need to agree and create documentation for best practices for code style, preferred libraries in priority order, etc. There are many ways similar tasks are accomplished in this app, so arguments can be made for both sides.

js/views/message.js
aidanamavi: You guys are right, I'll change this back.

aidanamavi: Otherwise you get "TypeError: thisModel.get(...).set is not a function".

js/views/folder.js
jancborchardt: So what's the call on coding style here – plain JS or Underscore? Do we usually use Underscore elsewhere?

Not sure if @DeepDiver1975 wants to make the call, so you @AidanAmavi & @ChristophWurst as our JS pros should decide together. :)

js/mail.js
DeepDiver1975: Yes - as of today the notification check only returns new messages - but it should also return deleted messages and (if possible) messages where flags did change.

@ChristophWurst gave it a try in #763

js/mail.js
DeepDiver1975: > @DeepDiver1975 You would be willing to implement hundreds of lines of code from a new library over several lines of code. Then you call this elegant solution kung-fu in comparison?

yes - because the timers have the bad habit to consume cpu even if the browser/tab is not active. We shall avoid timers/loops (yes I'm aware that I'm using a timer myself - I want to remove that as well).

js/views/message.js
aidanamavi: I'm confused. This was deleted a while back and isn't showing up for me. Which commit ID is this?

js/views/message.js
aidanamavi: I'm confused. This was deleted a while back and isn't showing up for me. Which commit ID is this?

js/views/message.js
DeepDiver1975: This error is most probably related to other code changes in here. I can debug this if you like.
But calling the server is no the solution.

appinfo/routes.php
aidanamavi: If this is true, should I change toggleStar, the code I referenced for changing flags? In which case, why wasn't this caught before? How do I do a git blame again? ;)

js/send-mail.js
aidanamavi: The setMessageFlag didn't have a method for answered, and using Backbone functions get/set is inconsistent; sometimes TypeErrors are abound.

js/views/folder.js
aidanamavi: Probably more TypeError issues. Why is syncing with the IMAP server a - serious - issue?

js/views/folder.js
aidanamavi: > If this is true, should I change toggleStar, the code I referenced for changing flags? In which case, why wasn't this caught before? How do I do a git blame again? ;)

My earlier remark on this issue.

js/views/message.js
jancborchardt: It’s at https://github.com/owncloud/mail/pull/834/files#diff-df51d65ab371ac49786ad664ec2d53dfL155, in the latest state of your branch. Github only shows comments on current code.

js/views/message.js
jancborchardt: Same as above, latest state of your branch: https://github.com/owncloud/mail/pull/834/files#diff-df51d65ab371ac49786ad664ec2d53dfL169

appinfo/routes.php
jancborchardt: Yeah @DeepDiver1975 what’s the call here. ;) Is this relevant for this PR or should we just tackle this in a different issue?

jancborchardt:
@aidanamavi I fixed the style issues @ChristophWurst noted in https://github.com/owncloud/mail/pull/834#issuecomment-117792380

Regarding pull-to-refresh (or refreshing at all): How do I do that (on desktop)? Scrolling to refresh, what I would expect, does not seem to work. Neither does clicking and dragging with the cursor.

As a sidenote I get this issue in owncloud.log – not present in master:
```
"PHP","message":"Use of undefined constant count - assumed 'count'
at \/home\/jan\/owncloud\/apps\/mail\/lib\/service\/unifiedmailbox.php#65"
"method":"GET","url":"\/owncloud\/index.php\/apps\/mail\/accounts\/-1\/folders\/YWxsLWluYm94ZXM%3D\/messages?from=63&to=83"
```

ChristophWurst:
@jancborchardt this has been fixed on master (#839), this PR needs a rebase

jancborchardt:
Ah k. Then this seems fine as far as I tested except the pull-to-refresh question.

jancborchardt:
Ah k. Then this seems fine as far as I tested except the pull-to-refresh question.

@aidanamavi can you rebase?

@ChristophWurst @DeepDiver1975 which are the things left to fix which are _really_ relevant to this pull request? The stuff with POST and such seems to just be carried over from master and could be a separate concern. Let’s not grow this PR bigger.

js/views/folder.js
ChristophWurst: @jancborchardt  I'm definitely no JS pro ;-)

I agree with @aidanamavi, we should document best practices somewhere (wiki). There have been many scenarios where I wasn't sure which framework to use (e.g. ``$.each(..)`` or ``_.each(...)``) too.

DeepDiver1975:
> @ChristophWurst @DeepDiver1975 which are the things left to fix which are really relevant to this pull request? The stuff with POST and such seems to just be carried over from master and could be a separate concern. Let’s not grow this PR bigger.

I still totally disagree with the proposed mechanism to reload the message list now and then.
This is a pointless waste of network calls to the server, causes waiting times for the user and creates load on the owncloud server and the imap server.

We really should not do this.

@aidanamavi I have to apologize if my words are too hard and I'm sounding rude - this is really not my intention. I am very thankful that you are digging into this stuff - which is not the easiest! Thanks a lot!

jancborchardt:
Btw @aidanamavi as mentioned in https://github.com/owncloud/mail/pull/830#issuecomment-116647652 it would be better / would have been better if you split this giant PR up into separate ones. Having all this in one makes this a pain to review and also a pain for you. When so much new stuff (which is not necessarily related, unlike for the unified inbox for example) is introduced all at once, problems are bound to happen.

js/views/message.js
aidanamavi: That is not the latest. Go to the commits tab of this pull, and you can see the latest state/commit.
https://github.com/owncloud/mail/commit/d276088f99401bb14043cdd2ee09e85170fbda95

js/views/message.js
aidanamavi: No, please don't debug. Continue to point out issues without resolution. I'd love to waste more time. :)

js/views/folder.js
aidanamavi: I don't see a need for including underscore.

js/mail.js
aidanamavi: You don't include another library to solve a simple problem. If you need to stop the timer, set a bind. When the window is out of focus, stop timer. I was planning on doing this as well, even though the performance hit is so negligible, hence waiting. I wonder why you're not making a new issue of it, since I'm not doing anything that isn't *already* done, even by you nonetheless.

js/mail.js
aidanamavi: And failed. This new method works fine.

You're saying, yes, this works, but we prefer doing it our way, which at this point is only theory. If that's the case, keep your app. I'll fork. This bureaucracy is ridiculous.

js/views/message.js
aidanamavi: > That is not the latest. Go to the commits tab of this pull, and you can see the latest state/commit.
https://github.com/owncloud/mail/commit/d276088f99401bb14043cdd2ee09e85170fbda95

aidanamavi:
Because @DeepDiver1975 apparently has the final say, and the blind siding bureaucracy involved in pull requests, I'm going to stop actively contributing to this app. I'm going to fork and try things with a new approach. All my best to you guys. It's been a wonderful learning process.

@jancborchardt @ChristophWurst @clementhk Thanks for your help!

DeepDiver1975:
> Because @DeepDiver1975 apparently has the final say

I don't have the final say - just prove me wrong and I'll happily revert my opinions.

> and the blind siding bureaucracy involved in pull requests

Can you please elaborate on this?

> I'm going to stop actively contributing to this app.

To be honest: I would be a pity to see you leaving. What is driving you away at the end? 

Once again: Please don't get my comments personal - but I don't like the way things are changed in here - prove me wrong and I'll happily accept this approach.

jospoortvliet:
honestly I think the mistake was to not say as first reply to the PR: "please split it up".

Because now, parts which are reviewed and agreed upon are waiting for parts for which the approach is not yet agreed on, creating more rebasing work and delaying it all. That's frustrating and yeah, creates more work which you could call bureaucracy ;-)

My suggestion, @aidanamavi  - split it up. The parts which are done can go it quick, the parts still under discussion can continue to be debated... Saves you work of further rebasing, too.

aidanamavi:
> and the blind siding bureaucracy involved in pull requests

@jospoortvliet @DeepDiver1975 
The Calendar rework PR is at least 10 times as big, and still accepted as it is without splitting it into smaller PRs. That’s the precedent I’m following. Therefore, contribution guidelines are not enforced and other contributors, like myself, can’t expect to be bound by them. Furthermore, my original intention, as I expressed to you, was to fix the syncing issue, and noticed I could address scrolling and flags as well as they touched the same code. Other PR’s do this, for example, Unified Inbox, which Thomas set the precedent for as well.

DeepDiver1975:
> The Calendar rework PR is at least 10 times as big, and still accepted as it is without splitting it into smaller PRs.

Just because others make mistakes it does not justify :wink: 



> Other PR’s do this, for example, Unified Inbox, which Thomas set the precedent for as well.

Some fixes/enhancements do require bigger changes and thus the PR is getting bigger.
But we should try to keep PRs self contained and focussing on once issue - just like @jospoortvliet mentioned in his comment above.

DeepDiver1975:
Anyway - let's just not nitpick on this and try to get things sorted out. 

This is my overall plan:
- mailbox sync should be done within checkNotification (yes - adjusting the name might be reasonable)
- we should minimize calls to the server as much as possible
- we should rely on existing features and functionalities by using frameworks and libraries
- we should not use timers (I'll come up with as solution for my own timer as well)
- the js side of this app should step by step be migrated to fully be a backbone app

js/views/folder.js
jancborchardt: Ok, then that’s a call. @DeepDiver1975 @ChristophWurst what do you think about that? Not include underscore, or yes? @aidanamavi further arguments against underscore?

appinfo/routes.php
DeepDiver1975: The very high level design principle is:
- use GET if you want to get anything from the server and the state on the server is not changed. GET is idempotent
- use POST if you want to change a state. POST is not idempotent
- use PUT if you want to set a state on the server. PUT is idempotent

jospoortvliet:
@aidanamavi dude, the calendar rework thing - how long did it take to get that merged? Oh, it started in July 2014 and is still not merged? Hmmm. If you are OK taking a year to get this PR merged then, sure, this is the way to do it ;-)

Seriously, though, you're right that there are bad examples. And we even documented that PR's should be as small as possible, see https://doc.owncloud.org/server/8.1/developer_manual/general/codingguidelines.html#general -> I admit that that is NOT there if you look at https://doc.owncloud.org/server/8.0/developer_manual/general/codingguidelines.html#general as this was a recent change, so - yes, we added that too late... ;-)

aidanamavi:
@jospoortvliet Good point, but the Unified Inbox point still stands. The only thing worth committing to a separate commit to was the infinite scroll. If it's any consolation, that was only several lines of code. Everything else was sync related, my original feature goal. I did read the 8.0, and read the 8.1 today as well. :)

I'm going to contribute to the #840 discussion to attempt to find common ground for the sake of a greater collaborative environment for future contributors.

js/views/message.js
jancborchardt: Same thing: https://github.com/owncloud/mail/commit/d276088f99401bb14043cdd2ee09e85170fbda95#diff-df51d65ab371ac49786ad664ec2d53dfR219

js/views/message.js
jancborchardt: > No, please don't debug. Continue to point out issues without resolution. I'd love to waste more time. :)

@aidanamavi we’re patient and helpful, and then these passive-aggressive comments. What’s the deal?

js/views/folder.js
aidanamavi: * Duplicates Javascript and jQuery functions.
* Doesn't always work as intended; has issues with Backbone scopes.
* Adds an additional learning curve.

appinfo/routes.php
aidanamavi: These functions change a state on the IMAP server. Why would that not qualify as a POST?

These two route's functions behave the same at toggleStar, which you wrote (git blame), and I used as references for these routes/functions.

ChristophWurst: It's about the second route, which shouldn't change anything on the server (``messages#getUnseenCount``).

js/views/folder.js
jancborchardt: But maybe this is something we should move to https://github.com/owncloud/mail/issues/840, the issue we created for that purpose? Otherwise the discussion will keep the PR back more and more.

jospoortvliet:
@aidanamavi I can't judge if this could have been split up - if not, then you are right and we'll just have to accept that getting something big like this merged is going to take time and effort. You should be commended for taking on such a big task and - well, you'll have to live with this being difficult ;-)

Thanks for staying constructive. Are you coming to the conf? Would be nice to get some hacking done there ;-)

appinfo/routes.php
aidanamavi: Ah, good catch. I'll change that.

jancborchardt:
I’m closing this for reasons mentioned above. TL;DR: It’s too large, tries to do too many things, fix too many issues. Should be split up into several commits and several pull requests.

PR #836, opened 2015-07-01
unified inbox: respect 'count' parameter and return only the first 'count' messages

### LINKED ISSUE #803: Remaining issues for unified inbox

ChristophWurst:
@DeepDiver1975 @aidanamavi @Xenopathic review please

partially fixes #803

### DIFF
--- a/lib/service/unifiedmailbox.php
+++ b/lib/service/unifiedmailbox.php
@@ -62,6 +62,10 @@ class UnifiedMailbox implements IMailBox {
 			return $a['dateInt'] < $b['dateInt'];
 		});
 
+		if (count >= 0) {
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
looks good - THX :+1:

clementhk:
LGTM

PR #838, opened 2015-07-01
Fix sending replies from unified inbox

DeepDiver1975:
@jancborchardt @ChristophWurst @aidanamavi please test - THX

### DIFF
--- a/lib/controller/accountscontroller.php
+++ b/lib/controller/accountscontroller.php
@@ -263,7 +263,12 @@ class AccountsController extends Controller {
 		if ($account instanceof UnifiedAccount) {
 			list($account, $folderId, $messageId) = $account->resolve($messageId);
 		}
-
+		if (!$account instanceof Account) {
+			return new JSONResponse(
+				['message' => 'Invalid account'],
+				Http::STATUS_BAD_REQUEST
+			);
+		}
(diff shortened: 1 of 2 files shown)


lib/controller/accountscontroller.php
```diff
@@ -252,22 +253,20 @@ public function create($accountName, $emailAddress, $password,
...
 		$bcc, $draftUID, $messageId, $attachments) {
 		$account = $this->accountService->find($this->currentUserId, $accountId);
-		if (!$account instanceof Account) {
-			return new JSONResponse(
-				array('message' => 'Invalid account'),
-				Http::STATUS_BAD_REQUEST
-			);
+		if ($account instanceof UnifiedAccount) {
+			list($account, $folderId, $messageId) = $account->resolve($messageId);
 		}
```
ChristophWurst: Could you please re-add the above ``if (!$account instanceof Account) {`` check like:

```PHP
if ($account instanceof UnifiedAccount) {
    list($account, $folderId, $messageId) = $account->resolve($messageId);
} elseif (!$account instanceof Account) {
   return ....
}
```
Otherwise requests could result into 500 errors

ChristophWurst:
Except for error handling, :+1:

Issue #848, opened 2015-07-09
dual scrollbar


mitpro:
At long e-mails got dual scrollbars (from iframe)

![owncloud_mail_dualscrollbar](https://cloud.githubusercontent.com/assets/12111420/8595065/48a32784-2647-11e5-857a-12322e017e78.PNG)

In file templates/index.php change attribute of iframe (scrolling) - class=mail-message-body: line 101 or 108:

iframe scrolling="no" src="{{htmlBodyUrl}}"   sandbox="allow-popups allow-same-origin" seamless

works well

![owncloud_mail_nodualscrollbar](https://cloud.githubusercontent.com/assets/12111420/8595061/460fc3d8-2647-11e5-9f82-84feecb0ba07.PNG)

owncloud version: 8.1 stable
php: 5.5.9
webserver: Apache/2.4.7 (Ubuntu)
server os: Ubuntu 14.04.2 LTS
client os: Windows 7 x64
browser: chrome & firefox

jancborchardt:
@mitpro can you submit a pull request with your fix? :) Then we can simply test it out and merge it if it works.

PR #849, opened 2015-07-09
update index.php - dual scrollbar

mitpro:
Dual scrollbar-problem (Issue #848)

### DIFF
--- a/templates/index.php
+++ b/templates/index.php
@@ -105,7 +105,7 @@ script('mail', 'jquery-visibility');
 		<div id="mail-content">
 			{{#if hasHtmlBody}}
 			<div class="icon-loading">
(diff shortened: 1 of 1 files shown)


jancborchardt:
@mitpro thanks! One problem with this is that in case the HTML mail is wider than the viewport (like some newsletters or notifications) the content is cut off and it’s not possible to scroll horizontally.

cc also @ChristophWurst @irgendwie @clementhk @zinks- for testing.

jancborchardt:
Moving to 0.3 as it’s not critical. @mitpro can you address my comment? :)

ChristophWurst:
closing this because of no activity

mitpro:
was searching for a solution that only the horitol scrollbar would be shown - but not realy relevant

PR #856, opened 2015-07-13
Using kwi/urllinker for link detection

### LINKED ISSUE #841: Wrong link detection
Non-HTML mails like the following cause problems:
```
Hi,
bla bla: http://example.com/

lorem ipsum (http://test.at/)
regards
```
This is shown as:
```HTML

			
			Hi,<br>
bla bla: <a href="http://example.com/<br ></a>
<br />
lorem" target="_blank" rel="noreferrer">http://example.com/<br>
<br>
lorem</a> ipsum (<a href="http://test.at/)<br ></a>
regards<br />
<br />
" target="_blank" rel="noreferrer">http://test.at/)<br>
regards<br>
<br>
</a>
			
		
```

DeepDiver1975:
fixes #841 

@jancborchardt

### DIFF
--- a/tests/service/htmltest.php
+++ b/tests/service/htmltest.php
@@ -13,6 +13,8 @@
 
 namespace OCA\Mail\Tests\Service;
 
+use OCA\Mail\Service\Html;
+
 class HtmlTest extends \PHPUnit_Framework_TestCase {
 
 	/**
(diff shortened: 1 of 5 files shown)


tests/service/htmltest.php
```diff
@@ -21,30 +23,29 @@ class HtmlTest extends \PHPUnit_Framework_TestCase {
...
+			['<a href="http://apps.owncloud.com/content/show.php/Music?content=160485" rel="noreferrer" target="_blank">apps.owncloud.com/content/show.php/Music</a>', 'http://apps.owncloud.com/content/show.php/Music?content=160485'],
+			['<a href="https://groups.google.com/forum/#!forum/ctpug" rel="noreferrer" target="_blank">groups.google.com/forum/</a>', 'https://groups.google.com/forum/#!forum/ctpug'],
+//			array('<a href="http://www.amazon.de/s/ref=nb_sb_noss?__mk_de_DE=%C3%85M%C3%85%C5%BD%C3%95%C3%91&url=search-alias%3Da/ps&field-keywords=Fax%2C+Kopierer+scanner+Laser&rh=i%3Aaps%2Ck%3AFax%5Cc+Kopierer+scanner+Laser">www.amazon.de/s/ref=nb_sb_noss?__mk_de_DE=%C3%85M%C3%85%C5%BD%C3%95%C3%91&url=search-alias%3Daps&field-keywords=Fax%2C+Kopierer+scanner+Laser&rh=i%3Aaps%2Ck%3AFax%5Cc+Kopierer+scanner+Laser</a>', 'http://www.amazon.de/s/ref=nb_sb_noss?__mk_de_DE=%C3%85M%C3%85%C5%BD%C3%95%C3%91&url=search-alias%3Daps&field-keywords=Fax%2C+Kopierer+scanner+Laser&rh=i%3Aaps%2Ck%3AFax%5Cc+Kopierer+scanner+Laser'),
```
clementhk: Is this not working?

DeepDiver1975: No - the generated link within the anchor is somehow broken - most probably due to some encoding-fu. I have to dig deeper. But even besides this - this approach is much more error prone.

clementhk:
By looking at the code, I think only domain will remain in the output, like
`http://oc.org?a=1&b=2` becomes `<a href="http://oc.org?a=1&amp;b=2" rel="noreferrer" target="_blank">oc.org</a>`
in https://bitbucket.org/kwi/urllinker/src/b5e097f03963454b4e8f32581265c8e57c3409e6/lib/UrlLinker.php?at=default#cl-129 , this is what we want?

clementhk:
Also, the test that you commented out did not consider url encode, like `&` should be `&amp;` in html, so I suspect that fail has something to do with it? Or the issue above.

clementhk:
Also, I tried digging into `ezyang/htmlpurifier`, gosh it is complicated.
It would be so much easier to add `noreferrer` when using `kwi/urllinker`...
But it doesn't seem to be possible unless we modify the code...

DeepDiver1975:
> By looking at the code, I think only domain will remain in the output, like

It will output domain and path - protocol and query will be removed

I'm fine with this adoption. @jancborchardt

clementhk:
One last thing (I think),

In this approach we are now encoding HTML code, I have not look through the whole logic but I don't think we should change it, or we should not change this behavior it in the same commit as fixing url parsing.
So `<p>test</p>` will now be `&lt;p&gt;test&lt;/p&gt;`. If this will only be used in text then it's ok.

DeepDiver1975:
> So <p>test</p> will now be &lt;p&gt;test&lt;/p&gt;. If this will only be used in text then it's ok.

the whole url detection is performed on plain text mails only - on html mails we simply rely on the fact than link are already properly declared

DeepDiver1975:
@clementhk does that make sense? THX for your feedback - very much welcome.

clementhk:
@DeepDiver1975 if we are using it only in plaintext, then the "inefficient" `TransformNoReferrer` would no longer be needed as all other `>` has became `&gt;`, so a simple `str_replace('>', ' rel="noreferrer" target="_blank">')` would be enough? (I know it's not as cool as yours :))

jancborchardt:
:+1: nice, links work again. @ChristophWurst @irgendwie double-check please. :)

PR #863, opened 2015-07-15
remove message from cached message list

ChristophWurst:
does only work if the message was deleted via the mail app

@jancborchardt


jancborchardt:
Nice! We still need to fix that for the unified inbox, but this at least fixes it for the specific ones. :+1:

PR #865, opened 2015-07-15
correctly ellipsize long account emails, fix #804

### LINKED ISSUE #804: account color indicator strange position
One of my mail account's mail address is pretty long, so an ellipsis was shown before the unified inbox merge. Now it's gone + the new color thing has a strage position with this specific account:

@jancborchardt @aidanamavi any idea?

jancborchardt:
fix #804, check it @ChristophWurst


ChristophWurst:
nice! :+1:

PR #869, opened 2015-07-15
do not include recipient in reply CC list

### LINKED ISSUE #846: Unified inbox: Reply autofill puts yourself in cc
Before sending a reply from the unified inbox you always need to remove your own email from cc, which is hugely annoying. Also, this causes the cc/bcc fields to be expanded by default for all mails and results in an amount of fluff we don’t always need to display.

cc @ChristophWurst @irgendwie @DeepDiver1975

ChristophWurst:
fixes #846

:beers:

### DIFF
--- a/lib/account.php
+++ b/lib/account.php
@@ -617,7 +617,7 @@ class Account implements IAccount {
 	}
 
 	/**
-	 * @return string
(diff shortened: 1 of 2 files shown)


jancborchardt:
:+1:

PR #871, opened 2015-07-15
fix title double dash for unified inbox, ref #803

jancborchardt:
Fixes the double dash as mentioned in #803 – please check @ChristophWurst ;D


ChristophWurst:
:+1:

PR #875, opened 2015-07-16
delete previous drafts if the currently composed message is an empty …

### LINKED ISSUE #735: Empty drafts are saved, if you type something and delete it again
Drafts were introduced here: #505 

**Problem:** If you compose a new message, type something, delete the text again and leave the composer somehow (click on a mail, folder etc.) an empty draft is saved.

**Expected behavior:** No draft is saved.

**Actual behavior:** An empty draft is saved.

ChristophWurst:
…message

partially fixes #735

This doesn't work with the unified inbox, because the DELETE route for messages doesn't handle the special ``-1`` account ID. I'll try to fix that in 0.3 by creating the unified inbox client-side (Backbone).

@irgendwie @jancborchardt review please :)

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -511,36 +511,72 @@ var Mail = {
 				draftUID: null
 			};
 			_.defaults(options, defaultOptions);
-			var url = OC.generateUrl('/apps/mail/accounts/{accountId}/draft', {accountId: accountId});
-			var data = {
-				type: 'POST',
-				success: function(data) {
-					if (options.draftUID !== null) {
-						// update UID in message list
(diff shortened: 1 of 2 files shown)


jancborchardt:
@zinks- @clementhk @PoPoutdoor @Gomez @irgendwie @colmoneill we need some reviews, homies – wake up! :)

clementhk:
the php part obviously looks good so frond-end guys :)

Gomez:
better than before. 
Still the message "Draft saved!" appears every time. 
 
@ChristophWurst "I'll try to fix that in 0.3 by creating the unified inbox client-side (Backbone)." Sounds like a big rewrite?

PR #876, opened 2015-07-16
Fix escaping issues - especially on > in message replies

DeepDiver1975:


jancborchardt:
cc @MacDada @clementhk cause this issue was caused by urllinker

jancborchardt:
Fixed, and links still work. :+1:

lib/service/html.php
```diff
@@ -30,7 +30,7 @@ class Html {
 	 */
 	public function convertLinks($data) {
 		$linker = new UrlLinker(true, false);
-		$data = $linker->linkUrlsAndEscapeHtml($data);
+		$data = $linker->linkUrlsInTrustedHtml($data);
```
MacDada: @DeepDiver1975 This will **not** escape HTML found in `$data`, but you use HTMLPurifier so it should be safe (as long as purifier is configured properly).

PR #881, opened 2015-07-16
[WIP] Add forward functionality

### LINKED ISSUE #95: add back »Forward« functionality
When clicking »Forward«, it should work similar to Thunderbird but more integrated:

* [ ] a new »Recipient / cc / subject / message« block opens below
* [ ] the message is appended in the message field, and people have room to type something above
* [ ] the button says »Forward«

ChristophWurst:
fixes #95

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -44,6 +44,8 @@ var Mail = {
 		var currentFolderId = null;
 		var currentAccountId = null;
 		var currentMessageId = null;
+		var currentMessageSubject = null;
+		var currentMessageBody = ''; 
 		var messagesLoading = null;
 		var messageLoading = null;
 
@@ -82,6 +84,22 @@ var Mail = {
 				},
 				set: function(newId) {
 					currentMessageId = newId;
+				}
+			},
+			currentMessageSubject: {
+				get: function() {
+					return currentMessageSubject;
+				},
+				set: function(subject) {
+					currentMessageSubject = subject;
+				}
+			},
+			currentMessageBody: {
+				get: function() {
+					return currentMessageBody;
+				},
+				set: function(body) {
+					currentMessageBody = body;
 				}
 			},
 			messagesLoading: {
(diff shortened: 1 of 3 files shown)


ChristophWurst:
@DeepDiver1975 @jancborchardt basic forwarding should already work. The only thing left is a full detailed header (from, to, subject, date). I would suggest to add the basic version in 0.2 and extend it's functionality in 0.3. What do you think?

jancborchardt:
Nice! Fully agree with doing the basics for 0.2 and enhancing for 0.3

I did some changes:
* correctly capitalize 'Fwd:' in subject
* add 3 empty lines above forwarded message to leave space for writing something, like we do for replies

I’d also like to position the Forward button. For that it would have to be inside the `<div class="composer-fields">` – at the same time not show in the New message composer of course. @ChristophWurst can you do that? If it’s too complicated we can do the positioning for 0.3 as well. It’s not too important.

:+1:

DeepDiver1975:
Hmm - I'm not yet happy with the mail body of the forwarded message.
Some text is html escaped and there are too many line brakes

DeepDiver1975:
forwarded html messages are pure junk.

Looks like we need an html capable editor one day?

jancborchardt:
Regarding the line breaks, let’s put in the exceptions we talked about, same as when formatting the previous email for inline replying:
* empty `div` or `p` should not create a line break
* `div` or `p` with just a `<br>` or `<br/>` inside should only create one line break instead of two
* probably some more things

We don’t need to have it perfect for 0.2, but reducing the faulty line breaks a bit would be good.

Yeah, about the full-HTML messages (like notifications) we need to care for in a future release … good enough for now though.

jancborchardt:
Again, we need some more reviews here @zinks- @clementhk @PoPoutdoor @Gomez @irgendwie @colmoneill :) Let’s get stuff running for 0.2

Gomez:
:+1: 
Totally good enough for 0.2! Great work @ChristophWurst :beers: 

Lets create a new ticket with follow up stuff..

Issue #880, opened 2015-07-16
Unified inbox caching issue


jancborchardt:
The unified inbox shows outdated data on every refresh or folder change. cc @ChristophWurst @irgendwie @zinks- @clementhk

jancborchardt:
Exact reproduction steps:
1. go to a specific inbox
2. delete a message
3. go to the unified inbox

The message will still show, until after the refresh is done. On every refresh though.

jancborchardt:
Fixed by https://github.com/owncloud/mail/pull/890

PR #890, opened 2015-07-20
replace message list cache on reload

ChristophWurst:
@jancborchardt please test if this fixes the caching errors you noticed (messages still being shown after they have been deleted)

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -546,6 +546,7 @@ var Mail = {
 			options = options || {};
 			var defaults = {
 				cache: false,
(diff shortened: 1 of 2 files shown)


ChristophWurst:
this should fix #880

jancborchardt:
Yup, seems to be fixed by this! :+1: 

Review please @DeepDiver1975 @irgendwie @LukasReschke let’s do it!

DeepDiver1975:
:+1:

PR #908, opened 2015-07-29
Fix prefetching messages in background marking them as read

### LINKED ISSUE #822: Prefetching messages in background marks them as read
As noticed by @DeepDiver1975, messages are marked as read once they are fetched in background

DeepDiver1975:


DeepDiver1975:
@jancborchardt as discussed

jancborchardt:
@Xenopathic @ChristophWurst @Gomez @rullzer this fixes the messages being marked as read when loaded in the background, fix https://github.com/owncloud/mail/issues/822 – please also review! :)

lib/controller/messagescontroller.php
```diff
@@ -274,13 +274,20 @@ public function saveAttachment($accountId, $folderId, $messageId, $attachmentId,
...
 	 * @return JSONResponse
 	 */
-	public function toggleStar($accountId, $folderId, $messageId, $starred) {
+	public function setFlags($accountId, $folderId, $messageId, $flags) {
 		$mailBox = $this->getFolder($accountId, $folderId);
 
-		$mailBox->setMessageFlag($messageId, Horde_Imap_Client::FLAG_FLAGGED, !$starred);
+		foreach($flags as $flag => $value) {
+			$value = filter_var($value, FILTER_VALIDATE_BOOLEAN);
+			if ($flag === 'unseen') {
```
ChristophWurst: can't we just send 'seen' to the server?

js/mail.js
```diff
@@ -1179,6 +1179,7 @@ var Mail = {
 					$('#forward-button').hide();
 				}
 
+//				Mail.UI.messageView.flagMessage(messageId, 'seen', true);
```
ChristophWurst: I guess this line can be removed ;-)

jancborchardt:
@DeepDiver1975 still got a message marked as read without opening it.

lib/controller/messagescontroller.php
DeepDiver1975: would require some more changes - but yes in a second pr ..

ChristophWurst:
@DeepDiver1975 @jancborchardt same here, some messages are still flagged as seen before I actually open them

jancborchardt:
@DeepDiver1975 to reproduce it, refresh the page 2–3 times. Then all mails will be marked as read.

ChristophWurst:
Digging into horde.log shows that when the message is fetched, peek mode is used. However, there's a second body fetch command that doesn't peek, hence the flag is set…

ChristophWurst:
Found it: https://github.com/owncloud/mail/blob/master/lib/message.php#L499

We need to peek there too

ChristophWurst:
@jancborchardt I added my fixes. Please test again :)

jancborchardt:
@DeepDiver1975 @irgendwie @PoPoutdoor @Gomez please also check! :)

Issue #585, opened 2015-05-04
Add jasmine Testing


Gomez:
With this pr i wanna add jasmine tests. And make the Build Automation work.

* [x] Manage js dependencies with NPM (development stuff)
* [x] Manage js dependencies for front-end packages with Bower 
* [x] Requirejs as file and module loader
* [x] Jasmine as test framework

I am not (yet) a js guy, so is this good to go? Choosed this tools, because they are known and don’t go away next week (and i work on other a project with it).
Are other tools used in ownCloud core, maybe? 

If this works well, i will move the js deps to this tools, too (jquery, backbone, handlebars ...)

Successor of https://github.com/owncloud/mail/pull/555

@DeepDiver1975 @jancborchardt

PR #917, opened 2015-07-31
[WIP] introduce requirejs

ChristophWurst:
I invested some time to split the current code base into AMD modules. With the help of require.js those modules can by loaded asynchronously and dynamically as needed.

From my point of view, the most important benefits are:
* decoupling code
* properly handle dependencies between modules and libraries
* work towards testable js code
* get rid of global variables

This WIP does already work without any issues as far as I tested it. However, there's still potential for further improvement, e.g. further decoupling of ``Mail`` by getting rid of stuff like ``Mail.State``, ``State`` could be it's own module. Currently there are many circular dependencies between ``Mail`` and ``Mail.XXX`` modules.

@DeepDiver1975 what's your opinion on that?

### DIFF
--- a/js/UI.js
+++ b/js/UI.js
@@ -1,0 +1,722 @@
+/* global Marionette, Handlebars, OC */
+
+/**
+ * ownCloud - Mail
+ *
+ * This file is licensed under the Affero General Public License version 3 or
+ * later. See the COPYING file.
+ *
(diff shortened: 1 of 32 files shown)


Gomez:
:+1: 
Great work! Needs testing but definitely a right step!

Gomez:
https://github.com/owncloud/mail/issues/585

Issue #929, opened 2015-08-09
message.js error with keyboard delete (tooltip issue)


PoPoutdoor:
If delete mail with mouse click on trash icon, no error.
After delete mail with keyboard, error in console:

> TypeError: Argument 1 of Document.elementFromPoint is not a finite floating-point value.

https://github.com/owncloud/mail/blob/master/js/views/message.js#L110

jancborchardt:
@irgendwie @ChristophWurst @zinks- any idea?

irgendwie:
@jancborchardt This happens bc a button event has no property clientX or clientY in comparison to mosue events. As currently the tipsies don't seem to work, is this workaround still needed (manually trigger mouseover?) or can we drop it?

jancborchardt:
The question is if we need the tooltips there. I’d say they are useful to display in the trash so you know that the deletion from there is permanent. (Not anywhere else though – just like before.)

Issue #923, opened 2015-08-06
Delete mail in the middle, next displayed mail is the one above before deletion


PoPoutdoor:
OC 8.1.0.8
Mail latest

Seems after you delete a message from the message list, default action is to load message newer than the one deleted.

Think about the procedure when you start working on message other than the first (top) one...

aspdye:
Please follow the issue template so we get the info we need to be able to debug and fix the problem:
https://raw.githubusercontent.com/owncloud/core/master/issue_template.md

jancborchardt:
@PoPoutdoor it is intended design to load the next newest mail. That allows you to work through mails chronologically. Especially helpful when there are multiple mails on the same topic while we do not have threading yet.

Or what do you mean?

PoPoutdoor:
Let's say I read mail from the top of list, I keep this.

I click somewhere on the list to read the message of interest, click trash icon while I'm done with that message.

Before the deletion, message below is what I expected to read next.
I repeat read-trash/keep until I'm done with mail.

With current default date sort, every time the next message to display is jump up. So, I have to click one message below to get my message read-trash.

Until date sort can be set at UI, this behaviour brings user bad impressions working with Mail.

jancborchardt:
Well, this is by design. If you want to read messages achronologically, you can start at the very top and delete from there, then the next one will be the next youngest. But if you want to properly read them chronologically, the next one will always be the next chronologically.

PR #936, opened 2015-08-13
Replace tipsy with tooltip

irgendwie:
As we moved from tipsy to tooltip (https://github.com/owncloud/core/pull/17075), I changed them for the mail app.

This resolves some issues / problems we had, like the recently https://github.com/owncloud/mail/issues/929, but also creates new problems:
- [x] weird transparency problems (the 'mail to' tooltips) @jancborchardt ideas?

Please review this and search for bugs! :+1:

### DIFF
--- a/js/views/composer.js
+++ b/js/views/composer.js
@@ -124,6 +124,8 @@ views.Composer = Backbone.View.extend({
 			replyCc: options.data.replyCc,
 			replyCcList: options.data.replyCcList
 		});
+		
+		$('.tipsy-mailto').tipsy({gravity:'n', live:true});
 
 		this.$el.html(html);
 
--- a/js/views/message.js
+++ b/js/views/message.js
@@ -2,9 +2,6 @@
 
 var views = views || {};
 
-$('.action.delete').tipsy({gravity:'e', live:true});
-$('.tipsy-mailto').tipsy({gravity:'n', live:true});
-
 views.Message = Backbone.Marionette.ItemView.extend({
 
 	template: "#mail-messages-template",
@@ -37,6 +34,8 @@ views.Message = Backbone.Marionette.ItemView.extend({
 				$(a).imageplaceholder(displayName, displayName);
 			});
 		}
+		
+		$('.action.delete').tipsy({gravity:'e', live:true});
 	},
 
 	toggleMessageStar: function(event) {
--- a/js/views/message.js
+++ b/js/views/message.js
@@ -91,7 +91,9 @@ views.Message = Backbone.Marionette.ItemView.extend({
 			}
 			// manually trigger mouseover event for current mouse position
 			// in order to create a tipsy for the next message if needed
-			$(document.elementFromPoint(event.clientX, event.clientY)).trigger('mouseover');
+			if(event.clientX) {
+			       $(document.elementFromPoint(event.clientX, event.clientY)).trigger('mouseover');
+			}
 		});
 
 		// really delete the message


jancborchardt:
I’m pretty sure the transparency problems come from using opacity on the element. The tooltip is in the container of the element, so instead of opacity we need to use text-color with alpha.

irgendwie:
@jancborchardt Awesome! Changed the container of the tooltip to body. :+1:
Should be ok now.

jancborchardt:
Ok now maybe squash your commits into one, to keep the history clean. :)

irgendwie:
Pfff...

jancborchardt:
:D no pff here! Keep stuff clean my friend

irgendwie:
done :stuck_out_tongue_winking_eye:

itheiss:
Ok. Here we go ....

1. The tooltip looks a bit odd. But I assume this is beause I am testing against 8.1 and Bootstrap tooltip is only available in 8.2, right? So this is no issue.

 tooltip

2. There is no tooltip when I move the mouse over the trash icon. But this was also an issue before (The trash icon to delete an account shows a tooltip for me?!?)

Although I can't reproduce #923 everything is working as expected :+1:

js/views/composer.js
```diff
@@ -124,6 +124,8 @@ views.Composer = Backbone.View.extend({
 			replyCc: options.data.replyCc,
 			replyCcList: options.data.replyCcList
 		});
+		
+		$('.tipsy-mailto').tooltip({placement:'bottom', container:'body'});
```
jancborchardt: While at it, can you rename the .tipsy-mailto class to `.mailto-tooltip`?

jancborchardt:
I would say to fix this and make it still compatible with ownCloud stable8.1 and older, there should be a check around the .tooltip call if it exists. If it does not exist, use .tipsy. @irgendwie Ok?

@Henni @PVince81 @MorrisJobke we also need to keep this in mind for other apps.

PVince81:
Well, or third party apps can simply import tipsy if they need it instead of relying on core bundling it.
Ideally apps should not rely too strongly on libraries shipped in core.
Else this would qualify as "breaking the API".

Another alternative is providing a tispy shim in core to ease the transition.

DeepDiver1975:
:-1: but this change will break the app on anything less then 8.2

There was the idea that only the implementation of the tipsy/tooltip foo changes in core - the "API" should stay the same - otherwise all apps break

DeepDiver1975:
> Another alternative is providing a tispy shim in core to ease the transition.

That shim was a requirement for me - not in there? :imp:

DeepDiver1975:
okay - here is the code I was looking for: https://github.com/owncloud/core/blob/56cff46dc32a0049840bb09f1769413eace064d6/core/js/js.js#L1846

No need to change anything in the mail app from my pov

Henni:
There are some problems with the shim as bootstrap tooltips don't support the live setting for example.

DeepDiver1975:
Then we pull thebootstrap tooltip into the app

jancborchardt:
@irgendwie any updates? :)

jancborchardt:
We should also fix this for 0.2 since the tooltips are broken. @irgendwie can you just copy the library from core into the app so it works on older ownCloud versions too?

irgendwie:
@jancborchardt Just copying the library would break the tooltips, because I would also need to import the stylesheet...

I now changed the position of tipsy creation, therefore they should work with both, using tipsy and the tooltip shim.
Also fixed #929 with the last commit.

Please test and give feedback!

jancborchardt:
Is it just that the tooltips are slightly transparent? I would say that is ok.

irgendwie:
@jancborchardt It's because of the container of the tooltip. With tipsy, there is no problem with the transparency. As I can't control the container while using the tooltip/tipsy shim, the transparency problem is back.

jancborchardt:
@irgendwie so apart from the minor transparency stuff this is now good to go?

ChristophWurst:
:+1: works!

jancborchardt:
Nice! :) Now onwards with the last stuff for 0.2

Issue #919, opened 2015-08-03
Clicking notification does not load the correct email


jancborchardt:
I think this was introduced with the caching and lazy loading. The problem is that when you click a notification, the Mail app is opened, but the latest messages are not yet loaded. Thus the mail from the notification can not be shown yet.

We should force load the relevant message in the detail view while still loading the message list I guess, that might fix it.

cc @DeepDiver1975 @ChristophWurst @irgendwie @Xenopathic

ChristophWurst:
Let's debug and this…

ChristophWurst:
There are some issues with caching, however, I don't think the 'correct email' is loaded automatically. Have a look at this: https://github.com/owncloud/mail/blob/master/js/mail.js#L291 – only the folder is opened, but not the message. This makes sense, as there might be multiple new messages and opening one of them would also flag it as seen.

jancborchardt:
> opening one of them would also flag it as seen.

Of course – but it would also be open then. :) That’s our behavior now, and we can further adjust it in other PRs to only peek, or when we have mark as read/unread.

ChristophWurst:
I would say opening one of the received messages is a new feature -> 0.3. I fixed some issues regarding caching in PR #937.

PR #937, opened 2015-08-14
cache new messages loaded in background

ChristophWurst:
This ensures messages are shown immediately when either clicking on a notification (which triggers a folder refresh and cached messages are shown first) or if you are in a folder other than the inbox with new messages and switch to the inbox (same reloading/caching issue).

possibly fixes issues reported in #919

@jancborchardt @irgendwie


jancborchardt:
:+1:  nice!

PR #951, opened 2015-08-24
round sender/receiver photos

jancborchardt:
This makes use of the core styles which might come with pull request https://github.com/owncloud/core/pull/18506 to make the profile photos round.

What do you think @ChristophWurst @irgendwie @zinks- @Xenopathic?


ChristophWurst:
Looks good if the avatar is a face. For any other contact images it might look odd…

PR #955, opened 2015-08-26
remove new uncached messages from folder list cache

### LINKED ISSUE #954: Initial load or folder switch shows outdated cache state
@DeepDiver1975 and I talked about this the other day and I’m sure you have seen this before @ChristophWurst @irgendwie @Gomez @PoPoutdoor 

1. Open the mail app
2. Delete some, read some, etc
3. Switch to another folder and switch back
4. Initially it still shows the list you initially saw when opening the app – not the one the client should know already cause you used it to delete and read mails.

I think we can consider this one of the main issues we should really fix for 0.2 – it feels very janky.

ChristophWurst:
I guess this issue was introduced with my PR #937. The problem was that messages (message bodies to be precise) that haven't been cached yet aren't removed from the cached folder message list. It's a scenario I didn't think of when writing this ``removeMessage`` function :see_no_evil: 

fixes #954

@jancborchardt @DeepDiver1975 test this please :)

### DIFF
--- a/js/mail.js
+++ b/js/mail.js
@@ -203,14 +203,14 @@ var Mail = {
 				if (message) {
 					// message exists in cache -> remove it
 					storage.remove(MessageCache.getMessagePath(accountId, folderId, messageId));
-					var messageList = this.getMessageList(accountId, folderId);
-					if (messageList) {
-						// message list is cached -> remove message from it
-						var newList = _.filter(messageList, function(message) {
-							return message.id !== messageId;
-						});
-						this.addMessageList(accountId, folderId, newList);
-					}
+				}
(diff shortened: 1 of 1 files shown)


jancborchardt:
@ChristophWurst can you check failed Travis?

I’m switching to the branch and will check if it works. :)

ChristophWurst:
> @ChristophWurst can you check failed Travis?
> 
> I’m switching to the branch and will check if it works. :)

I checked it already, but have no idea why it's failing. Seems like it's not this specific PR, other PRs also fail.

jancborchardt:
Nice, works well! :)

PR #957, opened 2015-08-26
don't show a placeholder if 'from' isn't set on the message

### LINKED ISSUE #952: Unified inbox problem
Currently the unified inbox doesn’t load, I get this console error with Firefox:
```
Using //@ to indicate sourceMappingURL pragmas is deprecated. Use //# instead jquery.min.js:1:0
Content Security Policy: Les paramètres de la page ont empêché le chargement d'une ressource à self (« script-src http://localhorst 'unsafe-eval' »). mail
Deprecation warning: tipsy is deprecated. Use tooltip instead. js.js:1903:2
Deprecation warning: tipsy is deprecated. Use tooltip instead. js.js:1903:2
Champs mot de passe présents sur une page non sécurisée (http://). Cela représente un risque de sécurité permettant le vol d'identifiants de connexion.[En savoir plus] <inconnu>
TypeError: seed is undefined placeholder.js:69:0
```

Any idea @ChristophWurst @irgendwie @PoPoutdoor @zinks- @Gomez?

ChristophWurst:
fixes #952 

If anybody has an idea why the message has no 'from' attribute, let me know ;-)

@jancborchardt please test if it's fixed now :-)

### DIFF
--- a/js/views/message.js
+++ b/js/views/message.js
@@ -30,10 +30,13 @@ views.Message = Backbone.Marionette.ItemView.extend({
 		this.setElement(this.$el);
 
 		var displayName = this.model.get('from');
-		_.each(this.$el.find('.avatar'), function(a) {
-			$(a).height('32px');
-			$(a).imageplaceholder(displayName, displayName);
-		});
+		// Don't show any placeholder if 'from' isn't set
+		if (!displayName) {
+			_.each(this.$el.find('.avatar'), function (a) {
+				$(a).height('32px');
+				$(a).imageplaceholder(displayName, displayName);
+			});
+		}
 	},
 
 	toggleMessageStar: function(event) {


jancborchardt:
Hm, I still get the same issue. :\

js/views/message.js
```diff
@@ -30,10 +30,13 @@ views.Message = Backbone.Marionette.ItemView.extend({
...
 		var displayName = this.model.get('from');
-		_.each(this.$el.find('.avatar'), function(a) {
-			$(a).height('32px');
-			$(a).imageplaceholder(displayName, displayName);
-		});
+		// Don't show any placeholder if 'from' isn't set
+		if (!displayName) {
+			_.each(this.$el.find('.avatar'), function (a) {
+				$(a).height('32px');
+				$(a).imageplaceholder(displayName, displayName);
```
jancborchardt: This being called still when there is no displayName set has got to be the issue, right?

jancborchardt:
Seems to work now :+1:

PR #958, opened 2015-08-26
change attachment template to underscore template

### LINKED ISSUE #844: Mailto composer template
The composer (opened if you click mailto links) doesn't render the template correctly:
bildschirmfoto von 2015-07-07 08 36 43

1. Attachment template does not show the actual file name, ``{{displayName}}`` is shown instead
2. Send button should be inside the composer's textarea

ChristophWurst:
For some mysterious reason, the mailto composer seems to use underscore to render the template while in contrast the standard composer uses Handlebars… :confused: 

This isn't necessarily a permanent fix, but I don't know how to resolve this now.

fixes #844

### DIFF
--- a/templates/compose.php
+++ b/templates/compose.php
@@ -7,7 +7,7 @@
 </script>
 
 <script id="mail-attachment-template" type="text/x-handlebars-template">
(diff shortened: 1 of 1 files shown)


jancborchardt:
Nice, works! :+1:

PR #960, opened 2015-08-27
show sender avatar again.

ChristophWurst:
Due to my little mistake in PR #957, no avatars are shown now. This PR brings them back again.

@jancborchardt @Gomez @irgendwie @itheiss easy to review ;-)


irgendwie:
Fixes the bug :+1: was wondering if the missing avatars are intended :smile:

PR #968, opened 2015-08-28
jscs plugins removed

Gomez:
plugins config are removed


Gomez:
@ChristophWurst We dont use it, so we can add it later. Produce an error with new jscs version.

ChristophWurst:
:+1:

PR #988, opened 2015-08-30
show pointer cursor when hovering message list as feedback

jancborchardt:
Your wish is my command @ChristophWurst


ChristophWurst:
nice! :+1:

PR #991, opened 2015-08-30
Block images

### LINKED ISSUE #424: Don't load external images
### Steps to reproduce
1. Go to mail app
2. Load any mail with linked external images

### Expected behaviour
The external images should not show up immediately, only after an additional request to do so (similar to Thunderbird's default behavior).

### Actual behaviour
External images are immediately loaded (proxied somehow via owncloud, but still).
This might be considered a security concern (as Junk mail senders get notified if their mail is read...).

### Server configuration
mostly not relevant
Mail app version: 0.1.3 (not sure if this is still the current one... according to master in github.com/owncloud/mail, it seems to be. it's not the current master version though but from some time ago).

### Client configuration
not relevant

### Logs
not relevant

jakobsack:
Fixes #424


irgendwie:
Nice work! :+1: LGTM

ChristophWurst:
Awesome!! :+1:

Gomez:
:+1: 
Nice work! 
@jancborchardt Maybe some design ideas? The blue is to blue?

jancborchardt:
@jakobsack I mentioned using a transparent PNG, can you please fix that?

PR #992, opened 2015-08-31
Adjust version to 0.3.0

itheiss:
I assume this also has to be 0.3.0 ...

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "Mail",
-  "version": "0.2.0",
+  "version": "0.3.0",
   "description": "ownCloud Mail App",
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Yes, thank you :+1:

jancborchardt:
It’s not 0.3.0 yet though? cc @DeepDiver1975?

ChristophWurst:
@jancborchardt we can't set it to ``0.3-alpha`` because some tools expect numeric versions numbers

jancborchardt:
Hm, ok. Then we probably should use something like 0.2.9 so we can use 0.3.0 when we actually release 0.3.0?

PR #1010, opened 2015-08-31
Improve image blocking

jakobsack:
This PR improves the blocking of images in HTML messages:

1. Tracking pixels (smaller than 5x5 px) will be replaced
2. Images attached to the mail will be displayed


Gomez:
There is a successor.

PR #1017, opened 2015-09-01
add jasmine/karma for js unit testing

ChristophWurst:
Tests don't pass, but the env is set up


ChristophWurst:
@Gomez I added unit tests to travis :grinning: Let's see if that really works…

Gomez:
:+1:

PR #1021, opened 2015-09-02
refactor folder loading to controller/service components

ChristophWurst:
Still not perfect, but another piece of code that shouldn't live in ``ui.js`` is now split into controller/service components.

@Gomez @irgendwie review please ;-)


Gomez:
:+1:

PR #1022, opened 2015-09-02
fix buttons for New message and Check messages

jancborchardt:
Please review @ChristophWurst @Gomez @zinks- @colmoneill 

* remove background from New message button to make it compatible with upcoming 8.2
* reduce whitespace to unified inbox a bit
* adjust margin so New message and Check messages button are aligned to each other regarding height
* fix text alignment / size of Check messages button

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -223,7 +223,7 @@
 #load-more-mail-messages {
 	display: none;
 	margin: 10px auto;
-	padding: 10px 10px 6px;
+	padding: 10px;
 }
 #load-more-mail-messages {
 	margin-bottom: 250px;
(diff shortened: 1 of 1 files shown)


ChristophWurst:
> adjust margin so New message and Check messages button are nicely aligned

nicely? It should be centered ;-)
bildschirmfoto von 2015-09-02 11-33-30

Other than that it looks good :+1:

ChristophWurst:
One thing I just notice: the ``+`` is like black while ``New message`` is grey. Should we maybe use the same color for both?

jancborchardt:
@ChristophWurst updated. The coloring will be fixed in a separate pull request where I also fix the icons :)

PR #1032, opened 2015-09-02
fix recursion, remove duplicates and add line breaks to translation e…

ChristophWurst:
…xtractor

@Gomez


Gomez:
:+1:

ChristophWurst:
> You’re all set—the translation-extractor-recursion-linebreak branch can be safely deleted.

:smile:

PR #1033, opened 2015-09-02
remove dots from buttons

ChristophWurst:
@jancborchardt as discussed

### DIFF
--- a/js/templates/message-list.html
+++ b/js/templates/message-list.html
@@ -1,7 +1,7 @@
-<input type="button" id="load-new-mail-messages" value="{{ t 'Check messages …' }}">
+<input type="button" id="load-new-mail-messages" value="{{ t 'Check messages' }}">
 <div id="emptycontent" style="display: none;">
 	<div class="icon-mail"></div>
 	<h2>{{ t 'No messages in this folder!' }}</h2>
 </div>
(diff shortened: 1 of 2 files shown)


itheiss:
What about the translation files e.g. l10n/de_DE.json? Will they be changed in a seperate PR?

ChristophWurst:
> What about the translation files e.g. l10n/de_DE.json? Will they be changed in a seperate PR?

Translation is handled via Transifex, therefore we don't change those files.

itheiss:
Oki! Does that mean "Check messages ..." and "Load more …" have to be changed directly in Transiflex to keep translations working? Still learning how everything is working ...

ChristophWurst:
> Oki! Does that mean "Check messages ..." and "Load more …" have to be changed directly in Transiflex to keep translations working?

Yes, exactly. That might unfortunately take some time…

> Still learning how everything is working ...

No worries ;-) There's also a decent documentation on ownCloud (app) development: https://doc.owncloud.org/server/8.2/developer_manual/

Gomez:
:+1:  Kill the dots!

jakobsack:
:+1:

jancborchardt:
Good detail improvement! :)

PR #1034, opened 2015-09-02
Make blocked images transparent as asked by Jan

jakobsack:
This replaces the current blue "blocked image" with a 1px transparent version.


Gomez:
Hmm, i don't like the transparent version. There should be a little difference, so i can see the place of the missing image?
A little bit grey? 
@jancborchardt

jancborchardt:
Now after using the image blocking for some time, it’s clear that we need to free the space which was occupied by the pictures. That is, there should be _no_ placeholder, with 0 width/height. Otherwise there’s a whole lot of useless space in most emails, and the layout is not really broken by just removing it.

jakobsack:
@jancborchardt @Gomez @ChristophWurst  Please have a look at the updated version, it hides the images ("display: none") and shows them when requested.

jancborchardt:
@jakobsack nice, much better! Thanks

PR #1035, opened 2015-09-02
Block backgrounds from style attributes

### LINKED ISSUE #1019: Block css background images in emails
Images defined in the style attribute should not be loaded unless the user asks for it (same behaviour as image tags)

jakobsack:
This commit fixes #1019

### DIFF
--- a/lib/service/htmlpurify/transformcssbackground.php
+++ b/lib/service/htmlpurify/transformcssbackground.php
@@ -1,0 +1,73 @@
+<?php
+ /**
+ * ownCloud - Mail app
+ *
+ * @author Jakob Sack
+ * @copyright 2015 Jakob Sack jakob@owncloud.org
+ *
+ * You should have received a copy of the GNU Affero General Public
+ * License along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+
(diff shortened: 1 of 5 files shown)


jakobsack:
@Gomez, @ChristophWurst: any comments?

Emails with background images are sent by google plus for example

lib/service/htmlpurify/transformcssbackground.php
```diff
@@ -0,0 +1,73 @@
...
+					$value);
+				return $name.':'.$value;
+			} else {
+				return $cssAttribute;
+			}
+		};
+
+		// Reassemble style
+		$cssAttributes = array_map($func, $cssAttributes);
+		$new_style = implode(';', $cssAttributes);
```
ChristophWurst: camel case please: ``$newStyle``

ChristophWurst:
Code looks good. I couldn't really test it because it seems like I don't have any mail with background images :/

@jancborchardt @itheiss @aspdye could you help testing please? :)

Gomez:
:+1: 
Works perfect! 
Tested with this code: http://backgrounds.cm/.
And the insert "HTML Option" in Thunderbird (if HTML Mail is enabled).

Issue #1025, opened 2015-09-02
Use loading views instead of manually adding classes with jQuery


ChristophWurst:
Currently loading spinners are shown by adding css classes to elements using jQuery. In certain situations this can lead to classes not being removed or added.

To solve that problem cleanly, I'd suggest we add 'loading views' which are shown until the 'real' view has finished loading.

An example would be loading accounts (left sidebar): Simply create a new view/template with only one element (div) that has the loading spinner class and add it to the DOM. Once the real view/data has been loaded, replace that loading view.

jancborchardt:
Sounds good, yeah.

ChristophWurst:
Moving as the LoadingView is not yet used in all places

PR #1046, opened 2015-09-04
use marionette view for rendering navigation

ChristophWurst:
no visual nor functional changes, just some cleanup for #1025 

@Gomez @irgendwie @jancborchardt @itheiss review please ;-)

### DIFF
--- a/templates/index.php
+++ b/templates/index.php
@@ -24,23 +24,21 @@ if ($debug) {
 	<div id="app-navigation" class="icon-loading">
 		<ul>
 			<li id="mail-new-message-fixed">
-			<input type="button" id="mail_new_message" class="icon-add"
-				value="<?php p($l->t('New message')); ?>" style="display: none">
+				<input type="button" id="mail_new_message" class="icon-add"
+				       value="<?php p($l->t('New message')); ?>" style="display: none">
 			</li>
-			<li id="mail-new-message-list">
-				<div id="folders"></div>
-			</li>
+			<li id="app-navigation-accounts"></li>
 		</ul>
 		<div id="app-settings">
 			<div id="app-settings-header">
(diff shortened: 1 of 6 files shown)


templates/index.php
```diff
@@ -9,26 +9,18 @@
...
 				<button class="settings-button"
 					data-apps-slide-toggle="#app-settings-content"></button>
 			</div>
-			<div id="app-settings-content"> </div>
+			<div id="app-settings-content"></div>
 		</div>
 	</div>
 	<div id="app-content">
-		<div id="mail_messages" class="icon-loading">
+		<div id="mail_messages">
```
itheiss: Is there a reason why some id's are dash separated and others are with underscore? It's would suggest to stick with the one or the other version ...

Regards

Ingo

ChristophWurst: > Is there a reason why some id's are dash separated and others are with underscore? It's would suggest to stick with the one or the other version ...
> 
> Regards
> 
> Ingo

I don't think so. We more or less stick to dashes AFAIK, but some 'old' id/classes using underscore have not been changed yet because changing could potentially breaking something (they are used both in JS and CSS). Therefore, if you add something new, use dashes ;-)

Gomez:
:+1: Done manual tests.

jancborchardt:
Needs to be rebased

Gomez:
Again a rebase :(

ChristophWurst:
Done.

@irgendwie @jakobsack @itheiss review please ;-)

Gomez:
:( Again rebase 
@irgendwie @jancborchardt @jakobsack Anyone so kind to test?

ChristophWurst:
@Gomez done.

Issue #981, opened 2015-08-30
Fix Makefile for packaging


ChristophWurst:
I just discovered that our Makefile doesn't really work (``dist`` is the default target, but it doesn't even exist).

Additionally, we'll have to add ``bower install`` to download all the JS dependencies before creating the package.

ChristophWurst:
We should also add requirejs optimizer to combine all JavaScript and template files (https://github.com/requirejs/text/issues/34)

ChristophWurst:
fixed in #1068

PR #1055, opened 2015-09-07
add requirejs optimizer for packaging

ChristophWurst:
basically this is groundwork for packaging the next release, see #981

@DeepDiver1975 @Gomez @irgendwie @jakobsack @jancborchardt  @zinks- 

With this PR we can optimize the packaged version of this app by putting all dependencies into a single file. This includes all the JavaScript and Template (HTML) files.

As running the optimizer is a task for packaging, I added the ``r.js`` call to our ``Makefile``. To separate between development and production environment, the only thing that came up my mind was the ``DEBUG`` setting of ownCloud. Therefore I use this variable to determine whether to include the optimized version or load any file asynchronously.
I still don't think that's the perfect, way… I'm sure someone will try to use the app with ``DEBUG`` set to true and no optimized file generated.

If anybody has an solution for the above problem, let me know ;-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ tests/clover.xml
 
 # Node.js modules
 node_modules/
+
+# optimized JS files
+js/mail.min.js
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,17 @@ all: dist
 clean:
 	rm -rf $(build_dir)
 
+dist: install-npm-deps install-bower-deps optimize-js
+
+install-npm-deps: package.json
+	npm install
+
+install-bower-deps: bower.json install-npm-deps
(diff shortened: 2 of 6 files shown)


ChristophWurst:
btw: when using the optimized version, about 55 fewer requests are needed for loading the app :rocket:

Gomez:
Cool. 
I dont have installed bower globally, the Makescript failed. Would be better if we only depend on npm and use the rest from the project dir?

ChristophWurst:
> Would be better if we only depend on npm and use the rest from the project dir?

Sure, makes sense. I updated my commit. Please test again ;-)

Gomez:
:+1:

PR #1063, opened 2015-09-09
Move composer install into Makescript

Gomez:
Now we have the complete dev-setup in the Makefile:

Just do:
```
sudo install node nodejs-legacy curl
make all
```
@ChristophWurst We have to document that debug mode needs to be on, for js development, right?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,11 @@ all: dist
 clean:
 	rm -rf $(build_dir)
 
-dist: install-npm-deps install-bower-deps optimize-js
+dist: install-composer-deps install-npm-deps install-bower-deps optimize-js
+
+install-composer-deps: composer.json
+	curl -sS https://getcomposer.org/installer | php
+	php composer.phar install
 
 install-npm-deps: package.json
 	npm install


ChristophWurst:
this basically fixes #981

Makefile
```diff
@@ -12,7 +12,11 @@ all: dist
 clean:
 	rm -rf $(build_dir)
 
-dist: install-npm-deps install-bower-deps optimize-js
+dist: install-composer-deps install-npm-deps install-bower-deps optimize-js
```
ChristophWurst: please rename ``dist`` to ``install-deps`` and remove the ``optimize-js`` rule from there

ChristophWurst:
From my understanding, the Makefile's only use case is packaging the app for the appstore/nightlies. Therefore, I'd say the default target should be ``appstore``

ChristophWurst:
Additionally, the ``appstore`` target should depend on ``clean install-deps optimize-js``

ChristophWurst:
We also have to update files excluded from the archive (bower.json, .bowerrc, translation-extractor.php,…). This can either be done in this PR or we create another once this one is merged.

Gomez:
I would use the Makefile for all kind of "destinations", like:
make release (==appstore)
make dev-setup (with phantomsjs ...)
make (default == nightlies)

Would use save many troubles with the different setups...

ChristophWurst:
> I would use the Makefile for all kind of "destinations", like:
> make release (==appstore)
> make dev-setup (with phantomsjs ...)
> make (default == nightlies)
> 
> Would use save many troubles with the different setups...

I'm not sure if that's a good idea. We would have to rely on ``make`` being available on every developer's system. Not all distros ship it by default…

There shouldn't be a difference between packaging nightlies and packaging for the appstore, right? :)

ChristophWurst:
I'm closing this for now because we already have that in our current Makefile.

ChristophWurst:
@Gomez do we need this branch or can we delete it?

Gomez:
No need. Deleted

PR #1068, opened 2015-09-11
fix makefile for packaging

ChristophWurst:
* optimizer issues fixed when executed outside owncloud app directory
* clean before packaging
* exclude any files that aren't needed
* works both for optimized (minified) version and debug version (all js files loaded asynchronously)

@enoch85 @Gomez @jancborchardt please test it like

```BASH
cd /tmp
git clone https://github.com/owncloud/mail
cd mail
make
```

Then you'll find the package in the ``build`` directory, which can be installed to an owncloud instance. Please also test if that works without any issues, thanks :-)

New issue created: composer install no longer pulls npm/bower dependencies as this would lead packaged dev dependencies.  @Gomez should we add that to the readme?


enoch85:
Tested and works without any errors :+1: 

I didn´t find any package in /build, but I think that´s because I didn´t patched the mail.tar.gz

Ok to merge.

enoch85:
Just for the record, this PR works like a charm! :+1: :+1:

PR #969, opened 2015-08-28
Fix message deletion by 'del' hotkey while adding a new mail account.

irgendwie:
Fixes #912

It's done by checking if the message list is visible or not. Good idea? Bad idea? Review pls! @ChristophWurst

Issue #1057, opened 2015-09-07
No search field


irgendwie:
Starting from owncloud 8.1, the search bar is missing for the mail app. Therefore search for mails is not possible anymore.

Maybe caused by by: owncloud/core#16810

Edit: This also happens for some other apps, like contacts, bookmarks, etc.
It works for example with the files app and the tasks app.

jancborchardt:
@DeepDiver1975 @ChristophWurst @Gomez any idea here?

DeepDiver1975:
This was for sure working some weeks back .....

ChristophWurst:
5fd5ab6c5e17fb43bcb7c7b9b379f8bcc76d1109 caused it. I must have forgotten something…

irgendwie:
Oh, you're right - didn't check for an old mail version >.<
So no 8.1 problem :+1:

ChristophWurst:
Ok, this is totally strange. The 'plugin' is registered like before, I even checked the passed object and it's the initialized 'Search' module (having both ``attach`` and ``filter`` method). I don't see why this doesn't work anymore.

@PVince81 since this issue came with introducing requirejs, could it be that there's a timing issue? I could think of something like we're 'too late', like the on-page-load-event has already been fired and no plugins can be registered afterwards. This is the only thing that I could think of at the moment since we're loading our modules asynchronously… There isn't really anything I changed apart from that.

Btw. ``attach`` and ``filter`` are never called on our search module.

irgendwie:
Seems to work again with the current master :+1:

Still **sometimes!** when I search and there are no results it shows
"No search results for {{searchTerm}}",
again sometimes the searchTerm is correctly replaced with my search term.
Ideas? @ChristophWurst

ChristophWurst:
@irgendwie when it does work, are you using the optimized js file or source files? (debug mode off or on)

ChristophWurst:
@PVince81 ping ^

ChristophWurst:
Seems to work if you use the minified js version. This supports my suspicion that this bug is a timing issue.

PR #1069, opened 2015-09-12
Fix message deletion hotkey if searchbar is focused.

irgendwie:
If you are in the searchbar and hit the 'del'-key, the current focused message gets deleted. With this commit if the searchbar is focused, messages won't get deleted.

**Also** re-adds the exception for the 'add mail account' dialog (#969), which accidentally got deleted in this commit:

https://github.com/owncloud/mail/commit/d7315a8522d574cdb728fcc18ae6453504b901ce#diff-e3f4efbe9f975868bcbac1c372abecf8L210

### DIFF
--- a/js/init.js
+++ b/js/init.js
@@ -213,8 +213,12 @@ define(function(require) {
 			switch (key) {
 				// If delete key is pressed:
 				case 46:
-					// If not composing a reply:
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Code looks good :+1: , but I can't test this because I still don't see the search field (#1057)

PR #1080, opened 2015-09-15
add more messagescontroller unit tests

ChristophWurst:
the controller is still not fully covered. For 100% coverage we'd need some refactoring, e.g. 'show' is a PITA to write unit tests for…

@DeepDiver1975 @Gomez @irgendwie @jakobsack @itheiss @zinks- @Xenopathic @clementhk


Gomez:
:+1:

PR #1083, opened 2015-09-15
increase timeout to prevent failing checks when travis takes too long

ChristophWurst:
@DeepDiver1975 @Gomez @irgendwie good idea? Today travis is again extremely slow, therefore all scrutinizer checks fail :/

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,9 +1,6 @@
 filter:
     excluded_paths:
         - 'vendor/*'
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Btw: who has permission to re-run scrutinizer checks if they time out (and thus fail)?

ChristophWurst:
How ironic. Scrutinizer timed out on this PR…

Gomez:
:+1 lets see if it helps...

PR #1084, opened 2015-09-15
enable dependency caching for travis to speed up CI builds

ChristophWurst:
* [x] cache composer packages
* [x] cache npm packages
* [x] cache bower packages

Directory | Size
---|---
node_modules | 163MB
js/vendor | 3.2MB
vendor | 12 MB

## Comparison of time needed to install dependencies

 | without cache | with cache
---|---|---
composer | 17.85s | 3.29s
npm | 67.09s | 40.50s
bower | 6.22s | 2.84s

I made this comparison by clearing the cache and running the test twice.

Result: we can save about 45s per build, that's 7.5min for running all 10 tests :rocket: 

@DeepDiver1975 @Gomez @Xenopathic @clementhk @zinks- review please

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,12 @@ branches:
     - master
     - /^stable\d+(\.\d+)?$/
 
+cache:
+  directories:
+    - $HOME/.composer/cache/files
+    - $HOME/.npm
+    - $HOME/.cache/bower
+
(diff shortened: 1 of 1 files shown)


Gomez:
:+1: Impressive work on your side and on travis side :)

PR #1090, opened 2015-09-16
use grey instead of transparency for quoted mails so they don't get unreadable

jancborchardt:
Previously the opacity was stacked, and hence quotes on multiple levels would become unreadable. This just makes all the quotes grey.

Please review @DeepDiver1975 @enoch85 @zinks-


ChristophWurst:
Awesome :+1:

PR #1093, opened 2015-09-16
use Makefile to set up dev environment

ChristophWurst:
@Gomez @enoch85 I think this is now what we agreed on: the Makefile can be used to install the dev environment via `make dev-setup`.

`make` or `make appstore` creates the package, just like it was before.

Feel free to take over this PR and adjust the info in the readme ;-)

@irgendwie @jancborchardt

This replaces #1063

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,14 +22,19 @@ install-deps: install-composer-deps install-npm-deps install-bower-deps
 install-composer-deps: composer.phar
 	php composer.phar install
 
-install-npm-deps: package.json
+install-npm-deps:
 	npm install --production
+
+install-npm-deps-dev:
+	npm install --deps
 
 install-bower-deps: bower.json install-npm-deps
 	./node_modules/bower/bin/bower install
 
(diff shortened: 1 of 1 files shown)


enoch85:
Fine by me :+1:

jancborchardt:
Works fine in fixing https://github.com/owncloud/mail/issues/1122 as per the commands @christophwurst mentioned, nice! Can we pull this in then or is it still work in progress?

jancborchardt:
Not sure if it’s completely related to this change here, but I get the problem again that sent mails cut the recipients name and only use the email address … this is quite annoying.

Also I get this error in the log:
```Class 'OCA\\Mail\\Horde_Mime_Headers_Date' not found at \/home\/jan\/owncloud\/apps\/mail\/lib\/account.php#203```

ChristophWurst:
> Can we pull this in then or is it still work in progress?

readme.md needs an update. I can add that later.

PR #1094, opened 2015-09-16
fix position of send/reply buttons, fix #484

### LINKED ISSUE #484: different 'Reply' button layout in FF/Chrome
The 'Reply' button is a little bit misplaced in FF

Currently it looks like this:

(left: Chrome, right: Firefox)

@jancborchardt I think this is something for you ;-)

jancborchardt:
It was fun while it lasted, but finally fix https://github.com/owncloud/mail/issues/484 and the layout in general looking strange.

Please review @ChristophWurst @colmoneill @itheiss

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -416,10 +416,8 @@ textarea.reply {
 	height: 100px;
 }
 input.submit-message {
-	position: absolute;
-	bottom: 5px;
-	right: 0;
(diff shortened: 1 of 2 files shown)


irgendwie:
Let's get that merged :+1: :smile:

ChristophWurst:
:+1: finally :-)

PR #1096, opened 2015-09-16
fix transparency of tooltip in mail header

jancborchardt:
The tooltips for the names in the mail header are now normal, not transparent anymore. Please review @irgendwie and everyone. :)


irgendwie:
:+1:

Issue #1091, opened 2015-09-16
Sending mails broken


jancborchardt:
Trying to reply to an email, I get the »Server error« notification. However my owncloud.log says:

```
[...]

PR #1097, opened 2015-09-17
split accountscontroller::send into testable chunks and add tests

ChristophWurst:
Since this is one of our most critical methods (used to send and reply messages), we should unit test this as much as possible to prevent issues like #1091.

I split the method into more testable chunks. Now the controller creates a new ``Message`` object (or ``ReplyMessage`` when replying) and passes it to the account, which handles the rest.

* [x] create message class
* [x] split method
* Unit tests
  * [x] Message
  * [x] ReplyMessage
  * [x] accountscontroller::send

There's further potential to improve new message and reply handling, but I'll try to keep this PR as small as possible and do everything step-by-step.

### DIFF
--- a/lib/account.php
+++ b/lib/account.php
@@ -1,7 +1,9 @@
 <?php
+
 /**
  * Copyright (c) 2012 Bart Visscher <bartv@thisnet.nl>
  * Copyright (c) 2014 Thomas Müller <deepdiver@owncloud.com>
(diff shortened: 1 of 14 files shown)


ChristophWurst:
Note to myself: once this is done: refactor accountscontroller::draft the same way (it's a copy of the send method)

Gomez:
What a PR! To much PHP Foo for my brain, but i tested the app hard and run all phpunit tests. All seems fine and better than before :) :+1:

PR #1101, opened 2015-09-17
exclude tests vom scrutinizer checks

ChristophWurst:
Since code coverage numbers generated by scrutinizer don't make any sense to me, maybe tests file shouldn't be analysed at all.


ChristophWurst:
nothing changed.

PR #1108, opened 2015-09-22
run CI js tests only once

### LINKED ISSUE #1105: [CI] test javascript only once
Currently javascript tests are executed for each php version and database type. Since it's the same every time (we don't depend on core or anything) it's useless to re-run those tests.

Running js tests only once would also speed up CI builds.

ChristophWurst:
fixes #1105

@DeepDiver1975 @Gomez @itheiss @irgendwie @Xenopathic does anybody have any objections of running JS tests only once?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ php:
 env:
   global:
     - CORE_BRANCH=master
+    - TEST_JS=FALSE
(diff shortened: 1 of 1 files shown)


ChristophWurst:
# Comparison
## Before (37min 55sec):
bildschirmfoto von 2015-09-22 11-10-36

## After (28min 13sec):
bildschirmfoto von 2015-09-22 11-12-47

clementhk:
Good idea, the test is taking longer and longer somehow.

ChristophWurst:
> the test is taking longer and longer somehow.

That *could* be the mail server. Maybe google slows intentionally if there are too many requests at once.

Gomez:
Good! Fixes #1105

PR #1115, opened 2015-09-25
update npm packages; update jshint ignore list

ChristophWurst:
This fixes warnings when running ``npm install``

Jshint ignore list has also been updated to the current file structure.


Gomez:
:+1:

PR #1058, opened 2015-09-07
load app in bootstrap.php to fix php unit tests

ChristophWurst:
this is needed in addition to #1023 because of new path restrictions of core (https://github.com/owncloud/core/pull/18839)

@Gomez @irgendwie @Xenopathic @nickvergessen @oparoz

PR #1117, opened 2015-09-29
Add missing implementation of ILogger::logException()

DeepDiver1975:


ChristophWurst:
Please also add this method to the unit tests: https://github.com/owncloud/mail/blob/master/tests/service/loggertest.php#L25-L36

Gomez:
Does we need to add the same version_compare?

https://github.com/owncloud/mail/pull/1058/files

We need to get this in before we can merge other stuff.

DeepDiver1975:
> Does we need to add the same version_compare?

yes - we need that on the unit test - will add ....

DeepDiver1975:
all green -> merge

Gomez:
perfect!

PR #1124, opened 2015-10-13
add missing use statement

ChristophWurst:
@jancborchardt please test if that resolves the issue you noticed (https://github.com/owncloud/mail/pull/1093#issuecomment-147550190)


jancborchardt:
Works fine :+1:

PR #1126, opened 2015-10-13
improve mail address handling

### LINKED ISSUE #1111: Replying to mail cuts off the name of recipient
When you reply to a mail from `"Ada Lovelace" <ada@example.org>` then the name will be cut off, only leaving the email address.

@ChristophWurst @irgendwie @Gomez to the rescue. ;)

ChristophWurst:
still far away from good code quality, but at least works better than before… Maybe I'll have another look at this smelly code later on.

@jancborchardt does this fix your reported issue?

fixes #1111

### DIFF
--- a/lib/account.php
+++ b/lib/account.php
@@ -20,6 +20,7 @@ use Horde_Mail_Rfc822_Address;
 use Horde_Mail_Transport;
 use Horde_Mail_Transport_Mail;
 use Horde_Mail_Transport_Smtphorde;
+use Horde_Mime_Headers_Date;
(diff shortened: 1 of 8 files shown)


jancborchardt:
Yep, works! :) Thank you :+1: 

What about the Travis builds? Please also review @DeepDiver1975 @Gomez @irgendwie

ChristophWurst:
> What about the Travis builds?

Sometimes the mail server denies the connection. Restarting the job fixed it ;-)

Issue #895, opened 2015-07-21
Error "String contains non-ASCII characters when sending emails


farson2003:
Whenever we send an email we get the message as in screenshot (OC8.1. with Mail 0.2)
![string](https://cloud.githubusercontent.com/assets/3358665/8795867/81a68b3a-2f88-11e5-8bb0-d960795e73bb.png)

DeepDiver1975:
Anything in data/owncloud.log?

farson2003:
Only:

Error mail Sending mail failed: String contains non-ASCII characters. 2015-07-21T08:09:41+00:00 
Error mail Saving draft failed: String contains non-ASCII characters. 2015-07-21T08:09:34+00:00 
Error PHP Automatically populating $HTTP_RAW_POST_DATA is deprecated and will be removed in a future version. To avoid this warning set 'always_populate_raw_post_data' to '-1' in php.ini and use the php://input stream instead. at Unknown#0 2015-07-21T08:05:23+00:00 


The last one I think is not relevant to you...

DeepDiver1975:
any idea which characters this might be? Any specific charset in place on your system? THX

farson2003:
Nothing special on the system!

clementhk:
Sounds like a horde mailbox name issue again?

clementhk:
Can you turn on debug mode and see what you got?
https://github.com/owncloud/mail#debug-mode

xn--nding-jua:
Hey,

I've commented the lines 69 to 76 within the file "apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/Data/Format/String.php"

```
        if (isset($this->_filter) &&
            $this->_filter->nonascii &&
            !($this instanceof Horde_Imap_Client_Data_Format_String_Support_Nonascii)) {
            throw new Horde_Imap_Client_Data_Format_Exception(
                'String contains non-ASCII characters.'
            );
        }
```

so no exception is thrown and the original error-message will be shown within the owncloud.log. Then you will see the original message "Sending mail failed: Could not open mailbox "Entw??rfe"."

As I'm using a German backend my "Draft"-folder is called "Entwürfe" and horde does not like this :)

Any suggestions?

xn--nding-jua:
I've done some more investigation. The problem occurs in this function-calling:

https://github.com/owncloud/mail/blob/master/lib/controller/accountscontroller.php#L441

Two lines before:
[...]

PR #1129, opened 2015-10-16
Fixed UTF7 encoding instead of UTF8 for mailbox-draftfolders

xn--nding-jua:
The constructor of "Horde_Imap_Client_Mailbox" requires a "false" for UTF-8 encoding. In the current code-base a "true" was set so that the horde-backend uses UTF-7 which caused problems with translated draft-folders (like "Entwürfe" which where mistransformed to "Entw??rfe").

### DIFF
--- a/lib/account.php
+++ b/lib/account.php
@@ -482,7 +482,7 @@ class Account implements IAccount {
 	public function deleteDraft($messageId) {
 		$draftsFolder = $this->getDraftsFolder();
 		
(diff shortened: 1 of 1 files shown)


clementhk:
Reviewed-by: me

Can someone test it?

ChristophWurst:
@farson2003 could you please test this? This should resolve your reported issue #895

PR #1133, opened 2015-10-18
update max-version to 9.0 so it works on current master

jancborchardt:
Please check @christophwurst @DeepDiver1975 @Gomez @zinks- @irgendwie :)


Gomez:
This is already done by @DeepDiver1975 some days ago.

PR #1136, opened 2015-10-20
Fix Debian/Ubuntu installation note in README.

jas4711:
This fix allowed me to install the dependencies properly on a Debian 8 machine.


ChristophWurst:
:+1: Good catch, thanks for fixing it ;-)

PR #1139, opened 2015-10-21
Ignore mail.min.js in jscs jshint check

Gomez:
Lets ignore the mail.min.js, confuse if you run grunt locally
@ChristophWurst Easy merge. No effect if its missing


DeepDiver1975:
:+1:

PR #1145, opened 2015-10-29
js coding guidelines

### LINKED ISSUE #840: Define coding & JS guidelines
We need to agree on some basic guidelines for our Mail app. Otherwise this will be confusing for new contributors. (Base for discussion is the comment from @DeepDiver1975 at https://github.com/owncloud/mail/pull/834#issuecomment-118004788) – FULLY DISCUSSABLE

1. mailbox sync should be done within checkNotification (yes - adjusting the name might be reasonable)
2. we should minimize calls to the server as much as possible
3. we should rely on existing features and functionalities by using frameworks and libraries
4. we should not use timers (I'll come up with as solution for my own timer as well)
5. the js side of this app should step by step be migrated to fully be a backbone app
6. (by aidan) which overlapping functionality should we use? i.e., typeof vs. isBoolean, or _.each vs. $.each, Javascript, vs. Underscore vs. jQuery vs. Backbone
7. ...

Please chime in @aidanamavi @ChristophWurst @irgendwie @clementhk @Xenopathic @LukasReschke

ChristophWurst:
@DeepDiver1975 @jancborchardt @irgendwie @Gomez please have a look at this. Is this what was meant in #840?

partly fixes #840

### DIFF
--- a/js/README.md
+++ b/js/README.md
@@ -17,3 +17,63 @@ inside the project's root directory. This combines and compresses all used JavaS
 files into one file: ``js/mail.min.js``. If [debug mode](https://doc.owncloud.org/server/8.2/developer_manual/general/devenv.html#enabling-debug-mode)
 is disabled, this compressed file is then used.
 
+## Coding guidelines
+
+Generally, any code contributed to this repository should comply with the general [ownCloud coding style guidelines](https://doc.owncloud.org/server/9.0/developer_manual/general/codingguidelines.html).
+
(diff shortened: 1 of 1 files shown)


Gomez:
Nice! Just add a new Todo 'Tests' (will fill as we go) and we should link to it from the main readme in the root folder.
:+1:

jancborchardt:
Nice! Merging already in developing state just to move fast and kick ass. ;)

PR #1147, opened 2015-10-30
fix search by adding a proxy

### LINKED ISSUE #1057: No search field

ChristophWurst:
I added a search proxy that is registered as app search on page load. The actual search module is then registered to that proxy shortly afterwards when it has been loaded asynchronously.

The only disadvantage of that approach is that the search field actually doesn't work for the time it needs to load the search module. However, that's actually a very short time on a production system since JavaScript source files are combined and compressed. Hence, search is not working for less than a second.

@jancborchardt @irgendwie @Gomez review please

fixes #1057

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -12,9 +12,6 @@ define(function(require) {
 	'use strict';
 
 	var timeoutID = null;
-	function attach(search) {
-		search.setFilter('mail', require('app').Search.filter);
-	}
 
 	function filter(query) {
 		window.clearTimeout(timeoutID);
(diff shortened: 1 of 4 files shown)


Gomez:
:+1: Works!

PR #1155, opened 2015-11-10
use h2 for redirect warning and correct heading size

jancborchardt:
Similar to https://github.com/owncloud/core/pull/20447, just use the h2 properly and fix the size so it’s like in core.

Please review @LukasReschke @owncloud/designers :)


jancborchardt:
Homies, give it a review :)

PoPoutdoor:
The warning text is more readable, feels better now.

PR #1173, opened 2015-11-26
fix unified inbox drafts

### LINKED ISSUE #1172: Unified inbox: error when saving drafts
Seems like our drafts functionality silently fails at the moment. The client sends ``-1`` as ``accountId`` which the server doesn't understand.

Possible solutions:

1. Fix it properly on the client side:
This requires the client to automatically save the draft to the currently selected account. Additionally, changing the account should also transfer the draft message to the newly selected account.

2. Add some server-side hacks as we did in other parts of the app:
The problem here is the decision of which account to save the draft to. The first one?

@jancborchardt should we add an error notification? If people rely on the draft functionality but it silently fails, this means loss of user data.

ChristophWurst:
fixes #1172

@owncloud/mail I need some testers for this. IMAP drafts are complex and I can't test all the different setups. Simply switch to this branch and see if there are any errors in the browser console. Thanks ;-)

### DIFF
--- a/js/views/composer.js
+++ b/js/views/composer.js
@@ -351,6 +351,11 @@ define(function(require) {
 			//  - undo lie - very important
 			//
 
+			// if available get account from drop-down list
+			if (this.$('.mail-account').length > 0) {
+				this.accountId = this.$('.mail-account').find(':selected').val();
+			}
+
 			// send the mail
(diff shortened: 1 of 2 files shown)


Gomez:
:+1:

PR #1181, opened 2015-12-10
Adding translation sync preproc command ...

### LINKED ISSUE #1031: Pre process task for transifex.com
As we lose the translations in the handlebar js templates, we created a script to add them back:
https://github.com/owncloud/mail/pull/1030

The script need to be added to transiflex. @jancborchardt Can you do this? Otherwise we wait for @DeepDiver1975

DeepDiver1975:
fixes #1031


ChristophWurst:
:+1:  Nice! This means we can also delete https://github.com/owncloud/mail/blob/master/translations.js

DeepDiver1975:
> :+1: Nice! This means we can also delete https://github.com/owncloud/mail/blob/master/translations.js

an be put to gitignore I assume?

DeepDiver1975:
Works - merging

ChristophWurst:
> an be put to gitignore I assume?

Exactly. Thanks!

PR #1183, opened 2015-12-22
Fewer global selectors

ChristophWurst:
This moves global selectors into the app-view, because it feels less hacky to me then. I also added some TODOs, where a bigger refactoring is needed.

Some day I would like to fully get rid of hacky files like init.js and ui.js. We should try to encapsulate all view code into view classes instead of global event handlers that nobody really remembers what they are doing :smile: 

@owncloud/mail have a look please. The easiest is to review each of the commits since I tried to do it step-by-step.

### DIFF
--- a/js/init.js
+++ b/js/init.js
@@ -53,10 +53,12 @@ define(function(require) {
 		};
 	})(), 1000);
 
+	// TODO: create marionette view and encapsulate events
 	$(document).on('click', '#forward-button', function() {
 		Mail.UI.openForwardComposer();
 	});
 
+	// TODO: create marionette view and encapsulate events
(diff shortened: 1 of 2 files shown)


ChristophWurst:
Failing tests are not related, this PR doesn't touch any PHP. See #1186

PoPoutdoor:
Seems I point to the wrong code, should be
https://github.com/owncloud/mail/blob/master/js/ui.js#L291

Other than this, :+1:

jancborchardt:
@Gomez @irgendwie @phsc @aspdye can we have another review? :)

Gomez:
:+1: Just a log (debug) that should be removed.

ChristophWurst:
Thanks for reviewing!

PR #1189, opened 2015-12-28
Add 'bulk mail' as junk/span special folder.

matiasdelellis:
Yahoo! mail use "Bulk mail" as junk/span and until now is considered a common folder.

### DIFF
--- a/lib/mailbox.php
+++ b/lib/mailbox.php
@@ -314,7 +314,7 @@ class Mailbox implements IMailBox {
 			'drafts'  => ['draft', 'drafts'],
 			'archive' => ['archive', 'archives'],
 			'trash'   => ['deleted messages', 'trash'],
-			'junk'    => ['junk', 'spam'],
+			'junk'    => ['junk', 'spam', 'bulk mail'],
 		];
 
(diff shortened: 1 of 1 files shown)


ChristophWurst:
:+1: Makes sense, thank you.

@owncloud/mail another review please

matiasdelellis:
Hi,
As a side note, the Spanish translations are wrong. Both "Trash" and "Junk" are translated as "Basura". Although it may be confuse, these directories are clearly different.

 * "Trash" => "Papelera" (On Yahoo or Gmail) or "Eliminados" (On Outlook)
 * "Spam" => "Spam" (On Yahoo or Gmail) or "Correo no deseado" (On Outlook)

I guess it's better respect proposed by gmail and yahoo.. What do you think about that? So then I propose in transifex

jancborchardt:
@matiasdelellis yeah, just change that in Transifex. Best change it to what is used in Gmail/Yahoo cause that’s our target audience.

jancborchardt:
@matiasdelellis and thank you for the contribution! :) Looking forward to further fixes from you, and you are very welcome in our IRC channel at #owncloud-mail (on freenode) as well.

PR #1200, opened 2016-01-02
replace deprecated Util::getServerHost, Util::getServerProtocol

### LINKED ISSUE #1199: Replace deprecated util methods
We're using some deprecated util methods of the owncloud core:
- [ ] ``Util::formatDate`` (not yet removed)
- [ ] ``Util::imagePath`` (not yet removed, https://github.com/owncloud/core/pull/21844)
- [x] ``Util::getServerProtocol`` (not yet removed)
- [x] ``Util::getServerHost`` (not yet removed)

ChristophWurst:
partially fixes #1199

### DIFF
--- a/lib/model/imapmessage.php
+++ b/lib/model/imapmessage.php
@@ -90,7 +90,8 @@ class IMAPMessage implements IMessage {
 		$this->htmlService = $htmlService;
 		if (is_null($htmlService)) {
 			$urlGenerator = \OC::$server->getURLGenerator();
-			$this->htmlService = new Html($urlGenerator);
+			$request = \OC::$server->getRequest();
+			$this->htmlService = new Html($urlGenerator, $request);
 		}
 
 		if ($fetch === null) {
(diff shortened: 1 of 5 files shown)


ChristophWurst:
This is yet another method that has been deprecated and removed, but the replacement is not available for all core versions we support: https://github.com/owncloud/core/blob/188acd5e4e4bf9e35e327ecf8389c75ead744732/lib/public/irequest.php#L162-L168

DeepDiver1975:
> This is yet another method that has been deprecated and removed

this is why travis is failing on oc7 and oc8 .... I see these options:
- keep the current code until we only support oc8.1+
- introduce a version dependent code base for these calls

ChristophWurst:
Ok, let's go for the easier one and replace those method calls later. Moving to 0.4

ChristophWurst:
I added a version-hack like in #1247

Gomez:
This needs a rebase. Lets run travis again..

ChristophWurst:
> This needs a rebase. Lets run travis again..

done

PR #1203, opened 2016-01-02
[WIP] check app code via occ command

ChristophWurst:
This will help us to detect usage of deprecated methods automatically.

Checks should pass as soon as #1199 is fixed.

---
This change is [Reviewable](https://reviewable.io/reviews/owncloud/mail/1203)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,6 +71,9 @@ script:
   # Test lint
   - find . -name \*.php -not -path './vendor/*' -exec php -l "{}" \;
 
+  # Run app code checker
+  - php ../../occ app:check mail
+
   # Run js unit tests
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Let's clean this up in 0.4

ChristophWurst:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/owncloud/mail/1203)*

ChristophWurst:
As discussed with @Gomez, this will not be possible soon as core does not provide a stable API yet and we'll need version-hacks if we support all supported owncloud versions.

PR #1222, opened 2016-01-14
remove cyclic dependency from Makefile

ChristophWurst:
This resolves the issues of unnecessarily installing composer dependencies when trying to optimize javascript files.

@owncloud/mail review please

This target is executed either directly or when packaging the app for the appstore, where dependencies are installed anyway.


clementhk:
LGTM

PR #1214, opened 2016-01-10
Create issue_template.md

aspdye:
according to #997

PR #1225, opened 2016-01-15
Aspdye issue template

ChristophWurst:
squashed version of #1214.

@owncloud/mail review & merge please

### DIFF
--- a/issue_template.md
+++ b/issue_template.md
@@ -1,0 +1,104 @@
+## ownCloud Core
+
+### Steps to reproduce
+1.
+2.
+3.
+
+### Expected behaviour
+Tell us what should happen
+
+### Actual behaviour
+Tell us what happens instead
+
+### Server configuration
+**Operating system**:
+
+**Web server:**
+
+**Database:**
+
+**PHP version:**
+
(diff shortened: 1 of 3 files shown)


issue_template.md
```diff
@@ -0,0 +1,104 @@
...
+sudo -u www-data php occ config:list system
+from within your ownCloud installation folder
+
+or 
+
+Insert your config.php content here
+(Without the database password, passwordsalt and secret)
+```
+
+**Are you using external storage, if yes which one:** local/smb/sftp/...
```
jancborchardt: This as well as encryption, external user-backend and LDAP as well seems irrelevant for us, no? Sure we need info, but we should also not make it a crazy chore to report issues.

ChristophWurst: True that! I'll remove it

jancborchardt:
In the log section down there we should probably also mention the horde.log again? https://github.com/owncloud/mail#debug-mode

jancborchardt:
Otherwise it’s fine :)

ChristophWurst:
I added the horde log section, so let's get this in.

Issue #1228, opened 2016-01-15
Tipsy of the "delete" button doesn't go away


ele0s:
*From #1223*
Hi,

As said in the issue cited above by @ChristophWurst : 
> > cons of folders are'nt shown and the black popup who was normaly displayed when the mouse is on the "delete" button is buggy/"always present" (I'm on Chrome) when we effectuate many operations on mails (For exemple, a lot of deletions)

> What do you mean? The tipsy does not go away?

Yes, the tipsy doesn't go away.
For more informations, I use the default theme

![capture du 2016-01-14 20-57-56](https://cloud.githubusercontent.com/assets/9059362/12335855/97c61886-bb01-11e5-8f05-052b62fdbf17.png)
![capture du 2016-01-14 20-58-06](https://cloud.githubusercontent.com/assets/9059362/12335856/989d9658-bb01-11e5-8c59-86b4aced002b.png)

ChristophWurst:
@owncloud/mail any idea?

jancborchardt:
@irgendwie you played around with this in the past, right? :)

ErikPel:
Has the tipsy plugin been disabled because of this or is there something wrong with my server/browser because I can't see the tooltips at all?

ChristophWurst:
> Has the tipsy plugin been disabled because of this or is there something wrong with my server/browser because I can't see the tooltips at all?

Tooltips are only shown before permanently deleting mails, hence you'll only find them in the trash-folder

ErikPel:
Can there be multiple tooltips shown at once without this bug or would it work if we just put (".tooltip").remove() into the delete function?
@ChristophWurst

jancborchardt:
Closing because https://github.com/owncloud/mail/pull/1250 fixes it, right? :)

PR #1250, opened 2016-01-30
Fix for #1228

ErikPel:


ChristophWurst:
From https://github.com/owncloud/mail/issues/1228#issuecomment-177316846:
> Can there be multiple tooltips shown at once without this bug or would it work if we just put (".tooltip").remove() into the delete function?

This could work as a temporary fix. Generally, it's error prone to have a jQuery selector executed on a class that might be reused in other parts of the application.
@ErikPel please also try 
```js
var _this = this;
this.$el.addClass('transparency').slideUp(function() {
    _this.$('.tipsy').remove();
    _this.$('.tooltip').remove();
```
That way the jQuery selector is scoped to the message view. That should be the way to go.

Thanks a lot for your contribution! You can join us in our IRC channel #owncloud-mail on freenode. If you need any help, let me know ;-)

ErikPel:
@ChristophWurst 
Is there reason why there are two different
`    $('.tipsy').remove();
`
functions ran in
https://github.com/owncloud/mail/blob/master/js/views/message.js#L84
and
https://github.com/owncloud/mail/blob/master/js/views/message.js#L89
?

I changed it the way you suggested @ChristophWurst 
Is it alright now?

ChristophWurst:
> @ChristophWurst
> Is there reason why there are two different
> $('.tipsy').remove();
> functions ran in
> https://github.com/owncloud/mail/blob/master/js/views/message.js#L84
> and
> https://github.com/owncloud/mail/blob/master/js/views/message.js#L89
> ?

That's indeed strange. I assume removing it once should be enough. I'll have a look…

ChristophWurst:
Nope, that does not work because the tipsy is actually placed outside the message element. Therefore ``this.$(...)`` doesn't catch it.
Please change it back to ``$(...)`` by removing your last commit. Sorry for pointing you to the wrong direction :see_no_evil:

ErikPel:
@ChristophWurst It's all good, man
Rebased to revert the change

ChristophWurst:
:+1: perfect. Thanks for you contribution. We'll merge this PR once 0.3 has been released!

PR #1176, opened 2015-11-27
fix mailto composer

ChristophWurst:
This brings back mailto link support. Works with recipient, cc, bcc, subject and body.

@owncloud/mail review please

fixes #1020

PR #1257, opened 2016-02-01
remove unused js/template file

ChristophWurst:
I forgot to remove those files in https://github.com/owncloud/mail/pull/1176/files

@owncloud/mail @ErikPel @Tagada85 review please


ErikPel:
:+1: 
Everything that worked before seems to work now.

Thought at first that those "Uncaught TypeError: Cannot read property 'currentTarget' of undefined" errors were because of this but then I switched back and I still get them

ErikPel:
:+1:  Not much to test here since those weren't used anyway

ChristophWurst:
> :+1: Not much to test here since those weren't used anyway

Exactly :)

PR #1258, opened 2016-02-01
decouple js modules

### LINKED ISSUE #944: JS refactoring/cleanup ideas for 0.4
I would like to share my thoughts and ideas on the JavaScript code enhancements I am planning for 0.3. To make it easy and painless to extend this app's functionality in the future, I personally think we have to clean up many parts of the current code base.

# Modularity and testability
As shown in my proof-of-concept/WIP PR #917 this app can be split into asynchronously loaded modules with relatively little effort. The use of require.js allows to split combined files like https://github.com/owncloud/mail/blob/master/js/views/attachment.js and create separate files for each model/view/whatever.
By using require.js for all components we can also get rid of global functions/objects. Additionally more JS code should become testable when refactoring components into modules with declared dependencies which can then be mocked for unit tests. #585

# Backbone JS + Marionette
As @DeepDiver1975 mentioned in https://github.com/owncloud/mail/issues/840#issuecomment-118425690 I would also like to use more Backbone models/views/collection in this app. This way we can eliminate some nasty hacks that are currently used to detect changed properties like https://github.com/owncloud/mail/issues/840#issuecomment-118425690. Backbone comes with a great event system that can be used for such tasks.
Furthermore a feature of Backbone collections I love is [sync](http://backbonejs.org/#Sync). It can be used to automatically persist client data changes to the server. This way we can remove some manually created AJAX requests because they are automatically created when models of the collection change.

# Consistent coding style for modules
Currently app modules (e.g. Mail.State, Mail.Background, Mail.Cache) are written using different coding patterns: object literal, module pattern, revealing module pattern and possibly another one. We should choose on of those (preferably the 'revealing module pattern') and change all other modules in the same way. This will make the js code easier to read and extend (especially for new contributors).

# Decoupled components
For better maintainability and testability it is useful when components (e.g. controllers, services) are loosely coupled. Small parts are already using a Marionette radio (the app is one) for communication. I would like to have all components talk to each other via one (or maybe a multiple) radios as mediators.

## Self-registering
Components should register all their events by themselves. Currently this is handled by the "app"-object which causes unnecessary dependencies.

@DeepDiver1975 @Gomez @jancborchardt @irgendwie what's your opinion on that? Any objections? Anything else we could/should improve? :-)

ChristophWurst:
partially fixes #944

@owncloud/mail @ErikPel @Tagada85 I could not find any errors when using this branch. Please review/test this

### DIFF
--- a/js/background.js
+++ b/js/background.js
@@ -78,13 +78,13 @@ define(function(require) {
 			// If it's okay let's create a notification
 			var tag = 'not-' + folder.accountId + '-' + folder.name;
 			var icon = OC.filePath('mail', 'img', 'mail-notification.png');
-			require('app').BackGround.showNotification(email, body, tag, icon, folder.accountId, folder.id);
+			showNotification(email, body, tag, icon, folder.accountId, folder.id);
 		}
 	}
 
 	function checkForNotifications() {
-		require('app').State.accounts.each(function(account) {
-			var localAccount = require('app').State.folderView.collection.get(account.get('accountId'));
+		require('state').accounts.each(function(account) {
+			var localAccount = require('state').folderView.collection.get(account.get('accountId'));
 			var folders = localAccount.get('folders');
 
 			$.ajax(
@@ -98,10 +98,10 @@ define(function(require) {
 					_.each(jsondata, function(f) {
 						// send notification
 						if (f.newUnReadCounter > 0) {
-							require('app').UI.changeFavicon(OC.filePath('mail', 'img', 'favicon-notification.png'));
+							require('ui').changeFavicon(OC.filePath('mail', 'img', 'favicon-notification.png'));
 							// only show one notification
-							if (require('app').State.accounts.length === 1 || account.get('accountId') === -1) {
-								require('app').BackGround.showMailNotification(localAccount.get('email'), f);
+							if (require('state').accounts.length === 1 || account.get('accountId') === -1) {
+								showMailNotification(localAccount.get('email'), f);
 							}
 						}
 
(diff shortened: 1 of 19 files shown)


ErikPel:
@ChristophWurst Sorry for another dumb questions but I'm new to this.

What am I supposed to do? Do I just install this branch to my server and play around with the stuff you changed or?

ChristophWurst:
> @ChristophWurst Sorry for another dumb questions but I'm new to this.
> What am I supposed to do? Do I just install this branch to my server and play around with the stuff you changed or?

Before merging any changes, we let other developers review the changes. Hence, I mentioned you so you can have a look at the code and give feedback if necessary ;-)

Do you have a local development/test instance? It's easier to install it there and you don't mess up your production system.

ErikPel:
@ChristophWurst Yeah I have personal vps I run owncloud in for this type of stuff.

I'll take a look when I get a change

js/background.js
```diff
@@ -14,6 +14,7 @@ define(function(require) {
 	'use strict';
 
 	var OC = require('OC');
+	var BackGround = this;
```
ErikPel: Seems like travis failed because you are not using _this(?)
This might be nitpicky but should background be written without the camelCase since it's just one word?

Also the scrutinizer lists 11 code style issues but they were there before too. Intentional?

ChristophWurst: > Seems like travis failed because you are not using _this(?)

Yes indeed. I'll have to fix that. I named it 'Background' since it's more or less a static class.

ChristophWurst:
@ErikPel The variable wasn't needed at all, I removed it.

ErikPel:
Everything that worked before seems to work now.

Thought at first that those "Uncaught TypeError: Cannot read property 'currentTarget' of undefined" errors were because of this but then I switched back and I still get them.

ChristophWurst:
> Everything that worked before seems to work now.

Thanks for reviewing!

> Thought at first that those "Uncaught TypeError: Cannot read property 'currentTarget' of undefined" errors were because of this but then I switched back and I still get them.

Strange. Please open a new issue for that error

> So all good other than those 8 scrutinizer code style issues but they were there before so we should probably fix them in separate commit. I can fix them after this get merged and do a pull request

Yes, let's fix those in 0.4. Thanks!

PR #1260, opened 2016-02-03
update excluded files for packaging

ChristophWurst:
I noticed that bower.json and the issue template are included in the 0.3 release


Gomez:
:+1:

PR #1292, opened 2016-02-16
Add imap docker for faster CI builds

### LINKED ISSUE #1291: Use IMAP docker on travis
According to the docs, it's possible to run a docker image on travis: https://docs.travis-ci.com/user/docker/

By using a docker image like [this](https://github.com/yorba/docker-dovecot-postfix) we could reduce the time needed for travis builds as IMAP tests are probably faster when running against a local server.

ChristophWurst:
fixes #1291

- Uses trusty beta instead of docker based infrastructure
  - it is needed for running docker images
  - MySQL has to be installed explicitly as it's not yet included
- Makefile is used for starting/managing the imap docker
- This can be used locally by all developers
- **phpunit now needs 2-3 seconds instead of 70** as the IMAP server is run locally and xdebug is disabled if not needed

## Comparison
Before (last master build):
bildschirmfoto von 2016-02-17 11-47-45
After (this PR):
bildschirmfoto von 2016-02-17 11-47-06
Summing up, it's 34:40 vs. 13:42. However, that's not all – the docker based version can also be run concurrently! 

@owncloud/mail have a look at this :rocket:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
-sudo: false
+sudo: required
+dist: trusty
 language: php
 php:
   - 5.4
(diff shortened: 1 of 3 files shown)


ChristophWurst:
This will also allow concurrent jobs on travis, which is an additional speed-up.

ChristophWurst:
Apparently caching directories is not (yet) possible on trusty images

Gomez:
Wow awesome!

DeepDiver1975:
nice!

PR #1299, opened 2016-02-17
Add bountysource badge

ChristophWurst:
As discussed with @Gomez on IRC, I created a bountysource org for this app: https://www.bountysource.com/teams/owncloud-mail


Gomez:
Thx!

Issue #900, opened 2015-07-21
Reply box hard to reach for long mails


Bullnados:
Why is the answer box at the bottom? If u have long Mails u have to scroll down every time.
Would it not be better to answer like in Outlook/Thunderbird. Perfect would be clicking answer/forward button and than reading mode change to writing mode. Would also more logical, because u read a mail form bot to top and not vice versa.

jancborchardt:
We should introduce a floating reply button at the bottom of the viewport if the reply box is below the edge of the viewport.

Bullnados:
But the box is still at bottom?

jancborchardt:
@Bullnados yes, as with basically all other email apps.

Issue #1252, opened 2016-02-01
Sending email with very long history


tobiasKaminsky:
The "send" button is on the bottom of the email edit field.
If the email contains a very long history the user has to scroll a very long way to send the mail.
(never tried if ctrl + enter is working).
The best would be that the send button is always visible, snapped to the bottom edge of the browser.

ChristophWurst:
cc @jancborchardt

jancborchardt:
This is basically a duplicate of https://github.com/owncloud/mail/issues/900 – and I’m working on it right now. :)

PR #1301, opened 2016-02-17
fix placement of reply button, ref #900

jancborchardt:
Fix https://github.com/owncloud/mail/issues/900 and https://github.com/owncloud/mail/issues/1252, cc @Bullnados @tobiasKaminsky for review.

What’s missing (cc @Gomez): When the reply button is disabled, clicking it should scroll down to the reply box and focus it as well.

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -422,8 +422,9 @@ textarea.reply {
 	height: 100px;
 }
 input.submit-message {
-	float: right;
-	margin-right: 0;
+	position: fixed;
+	bottom: 10px;
+	right: 15px;
 }
 #reply-msg {
 	float: left;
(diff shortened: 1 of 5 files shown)


Bullnados:
great :)
but atm i cant review it. Have lost focus on OC.

Gomez:
@ChristophWurst Events don't get fired on disabled inputs. I used this before, you know something better? http://stackoverflow.com/a/3100395/4046794

jancborchardt:
Any idea on this? JS wizardry needed @Henni @raimund-schluessler @MorrisJobke

Gomez:
Sorry @jancborchardt. I have done work here. Its just css wizardry now. Check it out and test it, i think you will see the problem.

PR #1302, opened 2016-02-17
reduce width of account indicator bar from 5 to 2px

### LINKED ISSUE #1241: Unified inbox: Account color indicators needed?
I was looking at a few other mail apps with unified inboxes, and most notably iOS Mail doesn’t do any indicator for the account.

This led me to think about if it’s really necessary to show this indicator. It’s a lot of color in addition to the already very colorful avatar placeholders, and it feels a bit much. The individual inbox view is nicer and cleaner.

What do you think @owncloud/mail?

jancborchardt:
Please review @owncloud/mail, fix #1241

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -178,7 +178,7 @@
 
 .mail-message-account-color {
 	position: absolute;
-	width: 5px;
+	width: 2px;
 	height: 69px;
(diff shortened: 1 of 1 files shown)


Gomez:
:+1:

Issue #1306, opened 2016-02-18
Consistent use of Marionette views


ChristophWurst:
While some parts of the client are already structured into Marionette views, some views are missing:
- a view for the content
- a view for the messages
  - contains the message listview (already exists)
  - a message (detail) view

I created a simple UML-diagram for the final design:

![view_classes](https://cloud.githubusercontent.com/assets/1374172/13155792/b08dd5d6-d67f-11e5-9223-15d6bdfee6e9.png)

All that should be possible with the different view-types Marionette provides, e.g. LayoutView, ListView, ItemView.

With those views in place it should finally be possible to kill ui.js and (re)move most of those jquery-hacks (e.g. replacing content views instead of showing and hiding different html elements).

@owncloud/mail @ErikPel @radlr any input is appreciated. Please let me know if there's anything we can improve before someone starts implementing.

Gomez:
As discussed with @jancborchardt We can drop the SettingsView and merge it with the three dot menu. Work is already done here.

Looks good! Which one can i implement?

ChristophWurst:
> As discussed with @jancborchardt We can drop the SettingsView and merge it with the three dot menu. Work is already done here.

Awesome! If it helps you can use at the sketch app for reference, it uses a three dot menu with marionette too: https://github.com/ChristophWurst/sketch/tree/master/js/view

> Which one can i implement?

I'll first have a look at the content and messagecontent view. That should be the foundation for any further works. So simply pick all the views you would like to create and let us know so we don't work on the same components ;-)

ChristophWurst:
Let's create a check list to track our progress:

View | Status | Developer
------|-------|---------
AppView | :white_check_mark:  | -
NavigationView | :white_check_mark: | -
NewMessageView |  | ?
AccountsView | :white_check_mark: | -
SettingsView | | @Gomez 
ContentView | | @ChristophWurst
SetupView | :white_check_mark: | -
MessageContentView | | ?
MessageListView | :white_check_mark: | -
MessageView | | ?
LoadingView | | ?

If anybody is working on a view, please update the table.

ChristophWurst:
> As discussed with @jancborchardt We can drop the SettingsView and merge it with the three dot menu. Work is already done here.

I thougt about that and I'm afraid I have to disagree. Removing the settings is fine, but putting the button for adding an account to existing account's menus is strange. I assume users won't expect to find it there. Can we leave that in the settings view until we find an appropriate place to put it?

[...]

PR #1309, opened 2016-02-20
Create MessageContentView, MessageView

### LINKED ISSUE #1273: Create 'ContentView'
As suggested by @Radlr we should create a content view (marionette view) to manage app contents (messages vs. setup). This will also help to implement #488.

ChristophWurst:
First step for #1306

TODO:
- [x] fix layout, css
- [ ] ~~create and add LoadingView~~ see https://github.com/owncloud/mail/pull/1309#issuecomment-186782163
- [x] fix reply and forward text
- [ ] ~~``TypeError: data is undefined, ui.js:262:7``~~ can't reproduce anymore
- [x] fix switching of setup and message content view

fixes #1273

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -570,6 +570,13 @@ input.submit-message {
 	overflow-y: auto;
 	overflow-x: hidden;
 }
+.mail-message-container {
+	position: absolute;
+	top: 0px;
+	right: 0px;
+	bottom: 0px;
+	left: 0px;
+}
 
 
 .new-message-attachments-action {
--- a/css/mail.css
+++ b/css/mail.css
@@ -4,6 +4,12 @@
 	top: 50%;
 }
 
+.container {
+	position: relative;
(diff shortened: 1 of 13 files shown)


ChristophWurst:
@Gomez I won't create/add the LoadingView in this PR as I realized it's a more work than I originally thought. Probably the easiest way to add it would be to throw away any of those ``$('…').addClass('icon-loading)`` and ``$('…').removeClass('icon-loading)`` and then add LoadingView where needed.
That should be done in a follow up PR. The only problem about that is that we'll have some glitches in the app until that PR is merged, as e.g. this PR will introduce some bugs regarding adding and removing the 'icon-loading' class. I hope that is okay for you.
cc @owncloud/mail

ChristophWurst:
@owncloud/mail is ready for review. Except for some tiny bugs with the loading icon (see comment above), everything works like before.

PR #1323, opened 2016-02-24
update js libraries

ChristophWurst:
:see_no_evil: https://www.versioneye.com/user/projects/567bd06ba7c90e003500033f :hear_no_evil: 

I did some testing and it seems that nothing broke.

The result can be seen here: https://www.versioneye.com/user/projects/56ce15e86b21e500355b10eb


ChristophWurst:
- [ ] ``GET http://localhost:8080/apps/mail/js/vendor/domready/ready.min.js``

ChristophWurst:
@owncloud/mail review please :-)

PR #1324, opened 2016-02-25
add missing favicon types

jancborchardt:
The .ico we need for some browsers and the -mask for Safari. Please review @owncloud/mail


ChristophWurst:
Is the favicon detected automatically? At least in code we set it to the png-version: https://github.com/owncloud/mail/blob/b7ca20984502edc6ed5c3eafd8598d2f5ce3f227/js/views/app.js#L76

We'll also need those two new favicon types for the notification favicon, right?

Which browser can be used to test this?

jancborchardt:
Not sure if the other browsers support the notification stuff. But yeah, that might be something to add.

At the moment I think it can be tested on IE9/10 and Safari.

ChristophWurst:
needs a rebase once #1325 has been merged

ChristophWurst:
needs a rebase

jancborchardt:
@ChristophWurst done

jancborchardt:
Let’s merge this @ChristophWurst @Gomez @irgendwie? :)

ChristophWurst:
@jancborchardt did you test it? :D If yes and it works, merge it!

ChristophWurst:
I tested with IE11. On the initial page load, the .ico-file is used as expected. However, switching tabs causes the js event to be triggered which overwrites the favicon with the .png-version: https://github.com/owncloud/mail/blob/master/js/background.js#L112

I'll have a look at this

ChristophWurst:
Apparently the same happens on Firefox. This PR actually fixes an unnoticed favicon-bug: on the initial page load, the core favicon is used instead of the mail favicon. Hence, let's get this in and fix the notification favicon in another PR.

Issue #1304, opened 2016-02-17
Mobile: reorder controls


jancborchardt:
Right now deletion of mails is only possible in the list. In the detail view, you can only reply.

I’m thinking of these controls for the detail view:
- back button top left, instead of the ability to open the sidebar
- delete top right
- maybe later also a star next to the title, but that clickable area could conflict with the back button. Something for later.
- send/reply floating bottom right (see https://github.com/owncloud/mail/pull/1301)
- full-width input fields, or at least improving the style a bit

PR #1332, opened 2016-03-06
don't hide date on mobile, list view exactly as on desktop

jancborchardt:
I found myself frequently missing the date of the mail when on mobile. There’s no real reason to hide it as usually there’s plenty of space right to the sender. So the elements are now just like on the desktop.

Please review @owncloud/designers @owncloud/mail

### DIFF
--- a/css/mobile.css
+++ b/css/mobile.css
@@ -59,17 +59,5 @@
 	width: 100%;
 }
 
-/* do not display date in message list because there is too little space */
-.date {
-	display: none;
-}
-/* more space for sender accordingly */
-.mail-message-summary-from {
-	width: 80%;
-}
-.mail-message-summary-subject {
-	width: 80%;
-}
-
 /* end of media query */
 }


ChristophWurst:
Awesome! I also found myself missing that important information.
christoph_rambo

ChristophWurst:
Btw. @jancborchardt what do you think about adding a screenshot of the mobile view to our awesome readme?

jancborchardt:
LOL THIS GIF! :D

jancborchardt:
@ChristophWurst I’d only do that when it’s more improved, see https://github.com/owncloud/mail/issues/1304

Also it’s fine to not advertise some awesome stuff because then people will just be pleasantly surprised. :)

ChristophWurst:
> Also it’s fine to not advertise some awesome stuff because then people will just be pleasantly surprised. :)

Okay, makes sense to me

Issue #1283, opened 2016-02-10
#1202 Could not resolve OCP\IDb! Class can not be instantiated


pierreozoux:
Hi!

I wanted to try the newly merged feature #1202  \o/
I used the nightly build of mail app + oC8.2 
is it compatible?
I guess not, I received an error on the command line
```
php ./occ 
An unhandled exception has been thrown:
exception 'OCP\AppFramework\QueryException' with message 'Could not resolve OCP\IDb! Class can not be instantiated' in /var/www/html/lib/private/appframework/utility/simplecontainer.php:87
Stack trace:
#0 /var/www/html/lib/private/appframework/utility/simplecontainer.php(106): OC\AppFramework\Utility\SimpleContainer->resolve('OCP\\IDb')
#1 /var/www/html/lib/private/appframework/utility/simplecontainer.php(66): OC\AppFramework\Utility\SimpleContainer->query('OCP\\IDb')
#2 /var/www/html/lib/private/appframework/utility/simplecontainer.php(85): OC\AppFramework\Utility\SimpleContainer->buildClass(Object(ReflectionClass))
#3 /var/www/html/lib/private/appframework/utility/simplecontainer.php(106): OC\AppFramework\Utility\SimpleContainer->resolve('OCA\\Mail\\Db\\Mai...')
#4 /var/www/html/lib/private/appframework/utility/simplecontainer.php(66): OC\AppFramework\Utility\SimpleContainer->query('OCA\\Mail\\Db\\Mai...')
#5 /var/www/html/lib/private/appframework/utility/simplecontainer.php(85): OC\AppFramework\Utility\SimpleContainer->buildClass(Object(ReflectionClass))
#6 /var/www/html/lib/private/appframework/utility/simplecontainer.php(106): OC\AppFramework\Utility\SimpleContainer->resolve('OCA\\Mail\\Servic...')
#7 /var/www/html/apps/mail/appinfo/register_command.php(14): OC\AppFramework\Utility\SimpleContainer->query('OCA\\Mail\\Servic...')
#8 /var/www/html/lib/private/console/application.php(68): require('/var/www/html/a...')
#9 /var/www/html/console.php(68): OC\Console\Application->loadCommands(Object(Symfony\Component\Console\Output\ConsoleOutput))
#10 /var/www/html/occ(11): require_once('/var/www/html/c...')
```

Could you help here?

Danke!

ChristophWurst:
> is it compatible?

It should be. I'll have a look at this

[...]

PR #1202, opened 2016-01-02
add console command for account creation

ChristophWurst:
This adds a simple command as described in #1125.

implements #1125 

TODO:
- [x] encrypt passwords
- [x] add unit test

@DeepDiver1975 @hitam4450 @owncloud/mail

PR #1353, opened 2016-03-17
Fix service resolution

ChristophWurst:
Fixes the bug where occ failes resolving our service as we're asking the wrong container for it.


Gomez:
Issue: #1202

Gomez:
Works perfect in 8.2 & 9 now. Fixes #1283 @pierreozoux would you retest?

PR #1364, opened 2016-03-23
add contacts app integration

quik87:
Integrate contact with mail, will allow in the new message view to use email in contcats data.


ChristophWurst:
@quik87 what is this PR about?

PR #1385, opened 2016-03-30
fix order in app navigation

jancborchardt:
The idea is to have the sorting start with: Files, Activity, Gallery, Mail, etc.

Please review @ChristophWurst @Gomez @DeepDiver1975


ChristophWurst:
:+1:

Issue #540, opened 2015-04-25
Show draft count


ChristophWurst:
Evolution shows a draft counter (similar to the unread message counter):
![](https://cloud.githubusercontent.com/assets/1374172/7332503/4a39c412-eb42-11e4-87f8-f7d5a6135f47.png)

I think this is great as you are reminded that there are messages left that have not been sent :)

jancborchardt:
Yup, we should have that! :)

tahaalibra:
currently...it gives the count for "unseen"...should this be updated to use "total"

tahaalibra:
or should there be both unseen and total..

ChristophWurst:
It should show the total number of messages, seen/unseen makes no sense for drafts imho.

jancborchardt:
Yeah, total number should be shown.

PR #1047, opened 2015-09-04
[WIP] Some bugfixes for the unread mail counter

irgendwie:

PR #1396, opened 2016-04-01
show total count for drafts folder

tahaalibra:
this solves issue #540 

Currently the drafts shows unseen/unread count. this solution would enable to show total count for the drafts folder only.

review @Gomez @ChristophWurst, This is my first PR, i am open for code review

### DIFF
--- a/js/templates/folder.html
+++ b/js/templates/folder.html
@@ -9,8 +9,8 @@
 	{{#if folders}}<button class="collapse"></button>{{/if}}
 	<a class="folder {{#if specialRole}} icon-{{specialRole}} svg{{/if}}">
 		{{name}}
-		{{#if unseen}}
-		<span class="utils">{{unseen}}</span>
+		{{#if count}}
+		<span class="utils">{{count}}</span>
 		{{/if}}
 	</a>
 	<ul>
@@ -23,8 +23,8 @@
 			<a class="folder
 				{{#if specialRole}} icon-{{specialRole}} svg{{/if}}">
 				{{name}}
-				{{#if unseen}}
-				<span class="utils">{{unseen}}</span>
+				{{#if count}}
+				<span class="utils">{{count}}</span>
 				{{/if}}
 			</a>
 			{{/each}}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## 0.4.2
  * Fix autocompletion (regression)
    [#1394](https://github.com/owncloud/mail/pull/1394) @ChristophWurst
+ * Fix show total email count for drafts folder
+   [#1396](https://github.com/owncloud/mail/pull/1396) @tahaalibra
 
 ## 0.4.1
  * Fix js error when adding attachments
--- a/js/views/folder.js
+++ b/js/views/folder.js
@@ -19,6 +19,17 @@ define(function(require) {
 
 	return Backbone.Marionette.ItemView.extend({
 		template: Handlebars.compile(FolderTemplate),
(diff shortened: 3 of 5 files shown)


js/templates/folder.html
```diff
@@ -10,7 +10,7 @@
 	<a class="folder {{#if specialRole}} icon-{{specialRole}} svg{{/if}}">
 		{{name}}
 		{{#if unseen}}
```
Gomez: We go in only if unseen. Drafts are always unseen. So you need to do further tests here.

js/views/helper.js
```diff
@@ -102,6 +102,13 @@ define(function(require) {
 		return t('mail', text);
 	});
 
+	Handlebars.registerHelper('printTotalOrUnseenCount', function(folder, unseen, total) {
+		if(folder == 'drafts') {
+			return total;
```
Gomez: jscs reports here: "requireSpaceAfterKeywords: One space required after "if" keyword at". You should install jscs and test your style.

Gomez:
Thanks for this PR! Please add a changelog enty, too.

ChristophWurst:
@tahaalibra thanks for that PR. I'd prefer to have that view logic inside the [folder view](https://github.com/owncloud/mail/blob/master/js/views/folder.js). That can be achieved easily by using [template helpers](https://github.com/marionettejs/backbone.marionette/blob/master/docs/marionette.view.md#viewtemplatehelpers) like we did [here](https://github.com/owncloud/mail/blob/master/js/views/composer.js#L25-L49).

Using a view helper is a valid approach, but it makes the code harder to maintain as you'd have the logic split into two separate locations. There's only one template that would make use of it.

So simply move the filter code over as template helper and fix the bug @Gomez mentioned, thanks!

tahaalibra:
Updated the code...Using template helpers seems much better option

ChristophWurst:
Thanks a lot, that looks very good!

We'll have to decrement that 'total' number when the user deletes a message. Otherwise the counter stays unchanged while deleting drafts. There's code that updates the 'unseen' value, so we just have to adjust that to also update the other value.
@tahaalibra we can either do that in this PR or a follow-up PR. Do you want to have a look at that?

ChristophWurst:
[This](https://github.com/owncloud/mail/blob/master/js/views/messagesitem.js#L82-L135) is where clicks on the trash icon are handled. You can use ``require('state').currentFolder`` to get a reference to the folder model, where the 'total' counter has to be decreased. Changes to that property *should* trigger a re-render of the folder view, but I'm not a 100% sure about that. One has to try ;-)

tahaalibra:
i think this also need to be updated for another case i.e if a user receives a new message...

tahaalibra:
Also the title bar still shows drafts with unseen....i will try to correct these problem...i will contact if i need any help

tahaalibra:
[x] the count on decreases if we delete a message in the drafts folder
[x] the title of the page now shows total count. this is for drafts folder only

one thing though, i checked this without this PR, the unseen count for the index folder does not if suppose the user has loaded the mail page, then he receives a new mail, then he clicks 'check messages'. the unseen message count does not get updated.

Gomez:
Thanks @tahaalibra! Works now. 

When i create further drafts, the counter does not reflect this immediately. A complete reload is needed. I will create a issue for it. 

The "unseen count for the inbox folder" is a known issue. There is a hacky (closed) PR #1047. Will create a further issue (as we dont have one).

Gomez:
@ChristophWurst And thanks for the explanation! Greatly appreciated!

Issue #1392, opened 2016-03-31
log SMTP communication


ChristophWurst:
Similar to the existing IMAP logging, SMTP logging is helpful to debug mail sending errors. Thankfully, horde already [integrates this](https://github.com/horde/horde/blob/master/framework/Mail/lib/Horde/Mail/Transport/Smtphorde.php#L69-L72), we just have to make use of it.

gouglhupf:
Thanks for committing my changes.

ChristophWurst:
@gouglhupf pro tip: if you add "fixes #1392" to the commit message or the PR description, github closes the referenced issue automatically ;-)

ChristophWurst:
@gouglhupf you are also very welcome to join us on our IRC channel #owncloud-mail on freenode ;-)

PR #1397, opened 2016-04-01
log SMTP communication #1392

gouglhupf:
I copied the if-clause form imap and changed it. After I try it I noticed it would be better to split the logs. So one for imap and one for smtp.


ChristophWurst:
Hi @gouglhupf,
thanks a lot for adding that feature. The changes look good, I'll give it a test. We should also update our readme and the issue_template as we're referencing horde.log there.

ChristophWurst:
:+1: awesome, this will help debugging SMTP errors a lot!

PR #1399, opened 2016-04-01
use LoadingView instead of manually adding/removing classes

### LINKED ISSUE #1025: Use loading views instead of manually adding classes with jQuery

ChristophWurst:
Now there's also a single spinner when loading a folder for the first time instead of two.

fixes #1025 

TODO:
 - [x] use LoadingView as a replacement for the NavigationView loading spinner
 - [ ] fix broken SettingsView (core's slide js magic does not work if the html of the settings menu is not present at page load)
 - [ ] add missing LoadingView before accounts have been loaded (https://github.com/owncloud/mail/pull/1399#issuecomment-204666363)

@jancborchardt @Gomez @tahaalibra could you try this branch and tell me if there's anything I forgot? I'm not 100% sure if the changes cover all possible transitions between the UI's states.

---
This change is [Reviewable](https://reviewable.io/reviews/owncloud/mail/1399)

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -18,6 +18,7 @@ define(function(require) {
 	// Load controllers/services
 	require('controller/accountcontroller');
 	require('controller/foldercontroller');
+	require('controller/messagecontroller');
 	require('service/accountservice');
 	require('service/folderservice');
 	require('notification');
--- a/js/app.js
+++ b/js/app.js
@@ -26,10 +26,14 @@ define(function(require) {
 	var Mail = new Marionette.Application();
 
 	Mail.on('start', function() {
+		Radio.ui.trigger('navigation:loading');
+		Radio.ui.trigger('content:loading');
 		Radio.account.trigger('load');
 	});
 
-	Mail.view = new AppView();
+	Mail.view = new AppView({
+		el: '#app'
+	});
 
 	return Mail;
 });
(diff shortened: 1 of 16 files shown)


jancborchardt:
Nice! Except one thing: Between the left navigation loading and the main content having loaded, there is only a spinner starting at a certain point. And if it loads from cache, seemingly not at all.

ChristophWurst:
@jancborchardt could you take a quick screenshot/video? I'm not sure if I understand which spinner you're referring to

jancborchardt:
There is no center area spinner while the left navigation is loading (in this case not fully yet) or already loaded:
capture du 2016-04-02 09-28-48

jancborchardt:
Shortly after this screenshot, the account list on the left is fully populated, and then all emails incl the current one appear. (From cache I assume)

ChristophWurst:
Thanks, I'll try to fix that too

Gomez:
Looks good. Lets clean it up and merge. After merging i do a 0.4.2.
@ChristophWurst Would you separate the commits?

PR #1403, opened 2016-04-06
fix min-height of new and reply textareas

jancborchardt:
This fixes the new message textarea being way too small.

Please review @Gomez @ChristophWurst @owncloud/designers


skjnldsv:
:+1:

PR #1411, opened 2016-04-12
Use LoadingView when loading messages

### LINKED ISSUE #1025: Use loading views instead of manually adding classes with jQuery

ChristophWurst:
fixes #1025 
@Gomez

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -18,6 +18,7 @@ define(function(require) {
 	// Load controllers/services
 	require('controller/accountcontroller');
 	require('controller/foldercontroller');
+	require('controller/messagecontroller');
 	require('service/accountservice');
 	require('service/folderservice');
 	require('notification');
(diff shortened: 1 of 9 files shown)


ChristophWurst:
@jancborchardt this should also fix the issue reported here https://github.com/owncloud/mail/pull/1399#issuecomment-204666363
the second commit should fix it – test please

jancborchardt:
Nice, works way better than the other PR now. :+1:

PR #1416, opened 2016-04-13
test IspDb with real xml data

ChristophWurst:
@Scheirle this is my attempt on testing xml based autoconfig automatically. What do you think? cc @Gomez @DeepDiver1975

### DIFF
--- a/tests/service/autoconfig/ispdbtest.php
+++ b/tests/service/autoconfig/ispdbtest.php
@@ -21,11 +21,10 @@
 
 namespace OCA\Mail\Tests\Service\Autoconfig;
 
-use OCA\Mail\AppInfo\Application;
 use OCA\Mail\Service\AutoConfig\IspDb;
-use PHPUnit_Framework_TestCase;
+use Test\TestCase;
 
-class IspDbtest extends PHPUnit_Framework_TestCase {
+class IspDbtest extends TestCase {
 
 	private $logger;
 
@@ -50,7 +49,7 @@ class IspDbtest extends PHPUnit_Framework_TestCase {
 	 * @param string $domain
 	 */
 	public function testQueryRealServers($domain) {
-		$ispDb = new IspDb($this->logger, Application::$ispUrls);
+		$ispDb = new IspDb($this->logger);
 		$result = $ispDb->query($domain);
 		$this->assertContainsIspData($result);
 	}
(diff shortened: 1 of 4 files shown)


tests/service/autoconfig/ispdbtest.php
```diff
@@ -50,12 +49,37 @@ public function queryData() {
...
+	public function testQueryRealServers($domain) {
+		$ispDb = new IspDb($this->logger, Application::$ispUrls);
+		$result = $ispDb->query($domain);
 		$this->assertContainsIspData($result);
 	}
 
+	public function fakeAutoconfigData() {
+		return [
+			['freenet.de', true],
+			//['example.com', false], //should it fail?
```
ChristophWurst: that does *not* fail, as we're not checking the ``emailProvider``. Is it redundant to check that?

Scheirle:
Disclaimer: I'm not (yet) very familiar with the owncloud code base, php and php unit tests.

The Application global [`ispUrls` variable](https://github.com/owncloud/mail/pull/1416/commits/fd48f7b389a712570c7cae5aa8cfe79c93d22fac#diff-98f41e4acfeac80f65d6397314947e94R53) feels out of place. It's such a specific setting for only one class. 

**Alternative Tests:**
- unit test for `IspDb::queryUrl(...)`
   1. Call `IspDb::queryUrl(dirname(__FILE__) . '/../../resources/autoconfig-freenet.xml')`
   2. Check that the file got parsed correctly (Is the imap hostname == `mx.freenet.de`, ...)
- unit test for `IspDb::query(...)`
   - Have a mock implementation for `IspDb::queryUrl` which reports back the requested url (at the end we have a list with all requested urls)
   - Have a mock implementation for `dns_get_record` which always return a predefined value.
   - Then we only have to check that the requested url list is the same as a predefined expected list.

Scheirle:
I began implementing what I mean: https://github.com/Scheirle/mail/commit/e4f631dd52c3a7b4b121aa049d4d676de6f92fda (currently not tested)

ChristophWurst:
@Scheirle I found an easier way: using the IHttpHelper to download the file and using ``libxml_load_string``. Then we can mock the httphelper interface and inject the xml file content. That's probably the easiest approach.

ChristophWurst:
@Scheirle I updated the PR

Scheirle:
@ChristophWurst Using `IClientService` sounds good. :+1: 

This should also allow us to easily fix #1417

ChristophWurst:
Damn, IClientService is not available on oc8 https://github.com/owncloud/core/blob/master/lib/public/http/client/iclientservice.php#L29

ChristophWurst:
@DeepDiver1975 what's the way to go here?

1. invoke a private method?
2. use a public getter method that can be mocked?

ChristophWurst:
@Scheirle @DeepDiver1975 ping

Scheirle:
@ChristophWurst using a mockable getUrls Methods sound fine, too.

PR #1426, opened 2016-04-17
Add Messageservice

ChristophWurst:
This adds an MessageService module, which is an abstraction of server communication like AccountService and FolderService.

Communication::get was not used anywhere, so I removed it.

@Gomez @irgendwie @skjnldsv @Scheirle @radlr


skjnldsv:
👍

PR #1431, opened 2016-04-17
Run/test appstore packaging on travis only once

ChristophWurst:
It's not necessary to run that on all builds. This will save some wasted CI power.


ChristophWurst:
@irgendwie @Scheirle easy to review

nickvergessen:
:+1:

PR #1432, opened 2016-04-17
Load next messages automatically when reaching end of the list

### LINKED ISSUE #499: Automatic lazy loading of more messages when scrolled to the end of the list

ChristophWurst:
fixes #499

@Gomez @jancborchardt @skjnldsv review please

### DIFF
--- a/js/controller/accountcontroller.js
+++ b/js/controller/accountcontroller.js
@@ -19,7 +19,7 @@ define(function(require) {
 	Radio.account.on('load', loadAccounts);
 
 	function startBackgroundChecks(accounts) {
-		setInterval(function(accounts) {
+		setInterval(function() {
 			require('background').checkForNotifications(accounts);
 		}, UPDATE_INTERVAL);
 	}
--- a/js/controller/foldercontroller.js
+++ b/js/controller/foldercontroller.js
@@ -1,11 +1,20 @@
 /**
+ * @author Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
  * ownCloud - Mail
  *
- * This file is licensed under the Affero General Public License version 3 or
- * later. See the COPYING file.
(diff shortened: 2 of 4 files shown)


jancborchardt:
Nice! :+1: 

Only thing is I see an error in the owncloud.log:
```
Argument 1 passed to OCA\\Mail\\Model\\IMAPMessage::hordeToAssoc()
must be an instance of Horde_Mail_Rfc822_Address, instance of Horde_Mail_Rfc822_Group given,
called in \/home\/jan\/owncloud\/apps\/mail\/lib\/model\/convertaddresses.php
on line 73
and defined at \/home\/jan\/owncloud\/apps\/mail\/lib\/model\/convertaddresses.php#33"
```

Also if we do lazy loading, we can remove the button from the bottom there and replace it with just a large centered spinner directly in the list.

ChristophWurst:
> Also if we do lazy loading, we can remove the button from the bottom there and replace it with just a large centered spinner directly in the list.

Yes, let's do that in a follow up PR ;-)

PR #1433, opened 2016-04-17
Increase Address column width to 256

### LINKED ISSUE #1421: Data too long for column 'email'
### Steps to reproduce
1. Answer on mail with a long return address.

### Expected behaviour
Email should be sent out.

### Actual behaviour
```Server Error``` appears at header.
Owncloud logs show an exception like this:

```
An exception occurred while executing 'INSERT INTO `oc_mail_collected_addresses`(`user_id`,`email`) VALUES(?,?)' with params [ 'id', 'my_mail_log_mail_address' ] right truncated: 1406 Data too long for column 'email' at row 1
```

### ownCloud Mail

**ownCloud Mail version:** (see ownCloud apps admin page)
0.4.1

#### Server configuration
**Operating system**:
docker image (9.0.1-fpm)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/32937297-data-too-long-for-column-email?utm_campaign=plugin&utm_content=tracker%2F216487&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F216487&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### PR DESCRIPTION
**Pull request by ChristophWurst**
fixes #1421

### DIFF
--- a/lib/controller/accountscontroller.php
+++ b/lib/controller/accountscontroller.php
@@ -25,6 +25,8 @@
 
 namespace OCA\Mail\Controller;
 
+use Exception;
+use Horde_Exception;
 use Horde_Imap_Client;
 use OCA\Mail\Account;
 use OCA\Mail\Db\MailAccount;
@@ -37,10 +39,11 @@ use OCA\Mail\Service\Logger;
 use OCA\Mail\Service\UnifiedAccount;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\JSONResponse;
 use OCP\AppFramework\Http\Response;
-use OCP\AppFramework\Http;
 use OCP\Files\File;
+use OCP\Files\Folder;
 use OCP\IL10N;
 use OCP\ILogger;
 use OCP\IRequest;
@@ -57,7 +60,7 @@ class AccountsController extends Controller {
 	/** @var AutoConfig */
 	private $autoConfig;
 
-	/** @var \OCP\Files\Folder */
+	/** @var Folder */
 	private $userFolder;
 
 	/** @var ILogger */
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by ChristophWurst** [comment]
@otbe @tahaalibra @Scheirle easy to review :-)

**Comment by jancborchardt** [comment]
@otbe @tahaalibra since you reported / were involved in the original issue, can you please review this?

File: lib/controller/accountscontroller.php [hunk 1]
```diff
@@ -25,6 +25,8 @@
 
 namespace OCA\Mail\Controller;
 
+use Exception;
```
tahaalibra: since we are using this...we can change \Exception $ex to Exception $ex on line 236

tests/controller/accountscontrollertest.php
```diff
@@ -203,17 +204,17 @@ public function testCreateAutoDetectFailure() {
...
 	public function newMessageDataProvider() {
 		return [
-			[false, false],
-			[true, false],
-			[false, true],
-			[true, true],
+			[false, false, false],
+			[true, false, false],
+			[false, true, false],
+			[true, true, true],
```
tahaalibra: looks good, just one question, do we need to check for other cases like [false, false, true] [true, true, false] etc.

ChristophWurst: Not really, but the tested method is so complex that I didn't want to write another test :see_no_evil: We really need to refactor it at some point.

lib/controller/accountscontroller.php
ChristophWurst: fixed. Thanks

viki53:
Are you sure 256 is a good idea? I tried editing it manually, but MySQL doesn't like `VARCHAR` to be over a size of `255` and I get an error…

ChristophWurst:
Hi @viki53,
you're right, we don't have enough people to test this app with all possible database configurations. You are very welcome to help us test released. Let me know and I'll ping you before the next release! Thanks.

Btw. this [has been fixed already](https://github.com/owncloud/mail/pull/1484). Keep in mind this app is still under development ;-)

viki53:
Okay, so when will this fix be pushed to the OC servers? Because in the meantime, my server is kinda dead. :-/ 

I'd be glad to test some apps, but I don't really use this one enough to test it (I just have it in case I'd need a webmail for some of my addresses some day).

ChristophWurst:
This is an open source project, *it's done when it's done* ;-)

It depends on which owncloud core version you're using. On newer ones, manually patching will make the code integrity check fail. So that might not be an option.
Apart from that, you could simply apply the fix (edit the database.xml file) and do the update again. I'm not totally sure if that works, but you could give it a try.

viki53:
I did try fixing that manually, but I still have the old file (with `64`) in place, not the new one.

The update triggers the error I mentioned earlier, so I guess it's not applied and the new files not saved.

PR #1434, opened 2016-04-18
Force visually distinguishable account colors

### LINKED ISSUE #1428: Force visually distinguishable colors in some cases
As a follow-up to the new color generator: https://github.com/owncloud/mail/pull/1425

There should be a way/function to force request visually distinguishable colors. For example when you need 2 or 3 different colors for mail accounts, calendars, etc. the colors should not necessarily be generated completely automatic. Otherwise you might end up with very similar colors (like me now for the mail accounts, and as other people reported with the old algorithm too).

cc @skjnldsv @owncloud/designers 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/33039035-force-visually-distinguishable-colors-in-some-cases?utm_campaign=plugin&utm_content=tracker%2F216487&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F216487&utm_medium=issues&utm_source=github).
</bountysource-plugin>

skjnldsv:
Divide the hue value by 20, round it and multiply it again by 20.
This allows for (360/20) 18 differents colors.

@jancborchardt 
Fix #1428

### DIFF
--- a/js/views/helper.js
+++ b/js/views/helper.js
@@ -36,7 +36,8 @@ define(function(require) {
 		var hash = md5(account);
 		if (typeof hash.toHsl === 'function') {
 			var hsl = hash.toHsl();
-			return new Handlebars.SafeString('hsl(' + hsl[0] + ', ' + hsl[1] + '%, ' + hsl[2] + '%)');
+			var hue = Math.round(hsl[0]/40)*40;
+			return new Handlebars.SafeString('hsl(' + hue + ', ' + hsl[1] + '%, ' + hsl[2] + '%)');
 		} else {
 			var maxRange = parseInt('ffffffffffffffffffffffffffffffff', 16);
 			var hue = parseInt(hash, 16) / maxRange * 256;


jancborchardt:
Hmm … either the value is too low or it’s not working. The colors are still very difficult to distinguish.

One is `hsl(140, 60%, 68%)` and the other `hsl(160, 60%, 68%)`. Maybe let’s increase the distance initially to 30 or 40? If there’s more, we can still fill up inbetween.

skjnldsv:
The calcul is simple, if we divide by 40, we get (360/40) 9 different colors! And it's just a random possibility, the chance we get 2 exact same colors are as high as any other one.
If you really want a different color we should not use the core generator.

Or we should allow the user to change the hash we're using to generate the color therefore allow him to change the color he wants (but we keep the randomisation)

jancborchardt:
The point is that the feel of the colors across ownCloud should be similar. So I would say we need to use the generator here.

Let's maybe increase it to 30 to see if that works better? Or have a separate function for multi-color request with a variable amount of parameters – and the function ensures they are as distinct as possible.

skjnldsv:
I feel like doing some terrible coding! :p 
I'm gonna do some screenshot of the different colors for 30 and 40 :)

skjnldsv:
Here you go: 20, 30 and 40.
contacts owncloud

jancborchardt:
Niice – yeah, that illustrates very well that we need the 40 value. With 30 still some green/turquoise and the rightmost eed values are difficult to distinguish.

skjnldsv:
Got it :)

skjnldsv:
Here you go! @jancborchardt !
But keep in mind that doesn't mean you will get different colors! You could very well get 3 times the same! There's a very poor chance, but still ;)

jancborchardt:
Yeah, this is better! :)

About the different colors, that’s why I suggested above:

> Or have a separate function for multi-color request with a variable amount of parameters – and the function ensures they are as distinct as possible.

And simply taking from the set of 9 above. That would be quite useful already for the multiple mail accounts and the calendar colors. :)

ChristophWurst:
tztztz… failing CI tests and @jancborchardt hits the merge button

jancborchardt:
I AM JUST MACHINE TAKING ORDERS


SEE MERGE BUTTON

HIT IT

skjnldsv:
ahahahahhaha

PR #1443, opened 2016-04-21
Make Logger independent of ILogger

ChristophWurst:
As this interface has now changed the second time, I changed our Logger to not explicitly implement that interface anymore. In fact, it's even impossible to implement because in 8.2 the method signature has the Exception typehint, which has been removed in 9.1 and PHP does not allow method overloads.

@DeepDiver1975 should this be considered when interfaces in core are changed?

@Gomez @Scheirle

### DIFF
--- a/lib/controller/accountscontroller.php
+++ b/lib/controller/accountscontroller.php
@@ -45,7 +45,6 @@ use OCP\AppFramework\Http\Response;
 use OCP\Files\File;
 use OCP\Files\Folder;
 use OCP\IL10N;
-use OCP\ILogger;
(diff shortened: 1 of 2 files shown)


ChristophWurst:
setting as priority as it blocks all other PRs

ChristophWurst:
@Scheirle does the :+1: reaction mean you're reviewd the PR? ;-)
If yes, please leave a comment instead because there is no notification for reactions

jancborchardt:
Can confirm Mail master doesn’t work on core master, but works with this pull request! :+1:

PR #1445, opened 2016-04-23
Make accounts collapsable and show only important folders

### LINKED ISSUE #677: Feature Request: Collapse account while using multiple accounts
I'm using mail with multiple mailaccounts. Because there are many folders it would be nice to collapse an account in the navigation bar on the left side to get a better overview.

Thanks for this great app!!

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/14824048-feature-request-collapse-account-while-using-multiple-accounts?utm_campaign=plugin&utm_content=tracker%2F216487&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F216487&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ChristophWurst:
fixes #677

Suggestions on what folders are to be considered important und should be shown even when the account is collapsed are welcome.

We can think about automatically hiding drafts if there are none, but as long as the client-side is not always up to date with the server state I'd like to not add that enhancement yet and wait until we have solved that cleanly.

@jancborchardt I'm afraid this is very badly discoverable. Any idea how to solve that?

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -170,6 +170,16 @@
 .mail-account-email:first-child {
 	padding-top: 50px;
 }
+.account-toggle-collapse {
+	display: block;
+	margin-top: 10px;
+	margin-left: 43px;
+	opacity: .2;
+	cursor: pointer;
+}
+.account-toggle-collapse:hover {
+	opacity: .5;
+}
 
 .mail-account-color {
 	display: inline-block;
--- a/js/templates/account.html
+++ b/js/templates/account.html
@@ -4,3 +4,6 @@
 <h2 class="mail-account-email" title="{{email}}">{{email}}</h2>
 <ul id="mail_folders" class="mail_folders with-icon" data-account_id="{{id}}">
 </ul>
+{{#unless isUnifiedInbox}}
+<span class="account-toggle-collapse">{{toggleCollapseMessage}}</span>
+{{/unless}}
--- a/js/views/account.js
+++ b/js/views/account.js
@@ -5,7 +5,7 @@
  * later. See the COPYING file.
  *
  * @author Christoph Wurst <christoph@winzerhof-wurst.at>
- * @copyright Christoph Wurst 2015
+ * @copyright Christoph Wurst 2015, 2016
  */
 
 define(function(require) {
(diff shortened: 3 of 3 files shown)


jancborchardt:
So far it doesn’t seem to collapse anything? Like Trash, Sent, Spam, all custom folders, (and maybe Drafts) should definitely be hidden by default.

ChristophWurst:
It works for me:

Collapsed state (default):
bildschirmfoto von 2016-04-24 12-05-49
Expanded state:
bildschirmfoto von 2016-04-24 12-05-58

Maybe it's a caching issue, try ctrl+F5

ChristophWurst:
Works on FF an Chromium

jancborchardt:
Arrrgh … again seemed to be a caching issue. ;) Works fine, and very nice! :)

Instead of clicking the account name to expand it (that should actually open the menu https://github.com/owncloud/mail/pull/1446), I would say there should be one more line saying »Show all folders …«. On the bottom, at 50% opacity which triggers the folders. Because that’s more related as the folders are shown below the important ones.

Then ideally there should also be a short animation to have some feedback of what’s happening, and the »Show all folders …« thing can change to »Collapse folders …«.

ChristophWurst:
@jancborchardt thanks a lot for your feedback. I updated the commit according to your suggestions: 
bildschirmfoto von 2016-04-24 13-25-19

An animation can be added later, I'll have to think about how to do that properly.

jancborchardt:
Nice! The line-height / padding is off a bit, that is between drafts and show all folders there's less space than between the othets. Is it a li as well? If not, should get the same padding values as the lis.

ChristophWurst:
Good catch, I fixed the css. No, it's not a li so I had to manually apply margin to align it with the other navigation entries.

ChristophWurst:
Looks better now:
bildschirmfoto von 2016-04-24 15-12-45

jancborchardt:
Nice, :+1:!

PR #1446, opened 2016-04-23
Add account popover menu and remove settings account section

### LINKED ISSUE #625: Integrate account settings better, move from settings section directly to sidebar
I’m thinking of something like this:
capture du 2015-05-06 12 56 42

What do you think @DeepDiver1975 @Xenopathic @wurstchristoph @owncloud/designers?

I’m definitely sure that the account settings (deletion, modification etc) should go directly there via the dot-dot-dot icon. I’m not entirely sure the »Add new mail account« function needs to go there as it’s not used that often.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/13293093-integrate-account-settings-better-move-from-settings-section-directly-to-sidebar?utm_campaign=plugin&utm_content=tracker%2F216487&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F216487&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ChristophWurst:
This kills the accounts section in the app settings and adds a popover menu to delete the account.

bildschirmfoto von 2016-04-23 18-11-00

@Gomez @tahaalibra @Tagada85 @skjnldsv @jancborchardt

~~If anybody remembers the issue please link it so it can be closed if this PR is merged, I cannot find it but I remember there was some discussion about it.~~
fixes #625


ChristophWurst:
Pro tip: to test this PR without actually deleting your account, you can add ``return;`` at the first line of the controller method [here](https://github.com/owncloud/mail/blob/c3c435bc4c1500619eec9dcd1a512e0fadc4f396/lib/controller/accountscontroller.php#L159) ;-)

jancborchardt:
We should have text there »Delete account« to fit with the actions menu in Files.

Also, did we say we want the add account function in that menu as well?

ChristophWurst:
> Also, did we say we want the add account function in that menu as well?

Could be, but I'm still against that. Sure, it would be cool to have no settings menu at all, but who would think you have to go to the menu of an existing account to create a new one? This action does not belong in there in my opinion.
Contacts does the same (button in the settings menu) and Calendar uses a separate button in the navigation. Also in files, you have the '+' button/action, you won't find that in the actions menu of existing files ;-)

Hence I'd say we should keep it as it is now until we have found a better place for the action.

ChristophWurst:
> We should have text there »Delete account« to fit with the actions menu in Files.

Could you please help me with that? I followed the [app css docs](https://doc.owncloud.org/server/9.1/developer_manual/app/css.html#menus) for the menu, but that does not tell what classes have to be used to allow text next to the icon. If I simply put text after the button element, the markup does crazy things :-)

ChristophWurst:
@jancborchardt just an idea, but how about adding popover actions to move accounts up and down?

Gomez:
@tahaalibra This would be a good GSOC start issue. You can take this over. Most of the work is already done here: https://github.com/ChristophWurst/sketch (good to see how it can be done).

jancborchardt:
@ChristophWurst yeah, separate thing though as it’s not that important either for now. :)

ChristophWurst:
- [ ] TODO: update changelog

ChristophWurst:
@tahaalibra is this ready to review?

tahaalibra:
@ChristophWurst you can review it..

ChristophWurst:
:+1: looks good and I assume you tested it too @tahaalibra

Gomez:
If i open the popover menu and click somewhere else, the menu stays. It should go away, like in the files app?

js/views/account.js
```diff
@@ -53,6 +65,45 @@ define(function(require) {
 		toggleCollapse: function() {
 			this.collapsed = !this.collapsed;
 			this.render();
+			this.listenTo(Radio.ui, 'document:click', function(event) {
```
ChristophWurst: @Gomez this code should catch that. @tahaalibra can you have a look at this please?

PR #1447, opened 2016-04-23
Remove ui.js and init.js

ChristophWurst:
Finally, those legacy js files can be removed :rocket: 

@Gomez @tahaalibra @irgendwie @DeepDiver1975 review please, this is needed for #1448


Gomez:
:+1: No errors found

PR #1455, opened 2016-04-25
move message communication.js methods to MessageService

ChristophWurst:
Moves remaining message related methods to the message service. Now ``communication.js`` is not necessary anymore as we have dedicated and loosely coupled services :rocket:

This does NOT fix any bugs regarding drafts that have been reported recently, this just moves code around.

- [x] Prefetching of messages in background still works
- [x] Sending messages still works
- [x] Saving drafts still works


Gomez:
:+1:

PR #1456, opened 2016-04-26
Update changelog to 043

Gomez:
late update for changelog


ChristophWurst:
PR is against master – intended?

ChristophWurst:
:+1: in case we need that on 0.4.3 as well we'd have to backport. But actually there is no need I guess

PR #1457, opened 2016-04-26
invalidate js cache if app version changes

### LINKED ISSUE #965: Cache invalidation for updates
Some changes like PR #956 require the cache to be cleared when you upgrade the app. Since we can't expect users to do that manually, some kind of cache invalidation should be used whenever the app version changes.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26275359-cache-invalidation-for-updates?utm_campaign=plugin&utm_content=tracker%2F216487&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F216487&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ChristophWurst:
fixes #965

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -21,7 +21,7 @@
     "domReady": "^2.0.1",
     "handlebars": "^4.0.5",
     "jquery": "2.2.3",
-    "jquery-storage-api": "jQuery-Storage-API#^1.8.1",
+    "jquery-storage-api": "julien-maurel/jQuery-Storage-API#~1.7.4",
     "jquery-visibility": "~1.0.11",
     "requirejs": "2.2.0",
     "text": "^2.0.14",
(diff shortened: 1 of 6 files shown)


Gomez:
First i get this:

```
ReferenceError: Cookies is not defined
<anonymous> jquery.storageapi.js:420
<anonymous> jquery.storageapi.js:23
<anonymous>
```

Later this:

callstack

I cleared localstorage, before.

ChristophWurst:
@Gomez good catch. I get those errors too after re-installing js dependencies.

ChristophWurst:
Downgrading the lib helps. We should check if we need that lib at all, it seems to be unmaintained

Gomez:
fine now!

PR #1461, opened 2016-04-28
Update mail.css

blackcrack:
Hi,
avoid overlap.

by the way, you should look with firefox
and use the "inspect/examine element" "Q"
there do you view many unneeded css-styles ..

best regards
Blacky

Changes:
´´
#mail-new-message-fixed {
	position: fixed !important;
	width: 241px !important; //beware of frames, avoid overlap !
---------------------------------------------------------
#mail-new-message-fixed {
    position: fixed !important;
    width: 230px !important;
´´

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -341,7 +341,7 @@
 /* editor */
 #mail-new-message-fixed {
 	position: fixed !important;
-	width: 241px !important;
+	width: 230px !important; //beware of frames, avoid overlap !
 	z-index: 1;
 	padding-bottom: 12px;
 	background-color: rgba(255, 255, 255, .75);


ChristophWurst:
Hi @blackcrack,
thanks for your patch. Which browsers did you test this? @jancborchardt and others tried to fix that, but we never found a solution that worked cross platform because the scrollbar width can not be determined by pure CSS. Our consensus was to fix that with JavaScript, but apparently nobody ever did that.

css/mail.css
```diff
@@ -341,7 +341,7 @@
 /* editor */
 #mail-new-message-fixed {
 	position: fixed !important;
-	width: 241px !important;
+	width: 230px !important; //beware of frames, avoid overlap !
```
ChristophWurst: no comment needed, please remove it.

ChristophWurst: btw. CSS uses ``/* comments */``

ChristophWurst:
> there do you view many unneeded css-styles ..

Pull requests to fix that are always welcome ;-)

blackcrack:
Hi Chris,

Firefox, actual 46.0 .
not all can fix with javascript, sometimes mus put the hand to the original things.. 
and by the way, in css it is able for setting up directly for different browsers, 
like that : 
background-image: -webkit-linear-gradient(top, #3498db, #2980b9);  //only for webkit
background-image: -moz-linear-gradient(top, #3498db, #2980b9);     // only for moz..illa
background-image: -ms-linear-gradient(top, #3498db, #2980b9);      // only for the old IE's M$ Browsers
background-image: -o-linear-gradient(top, #3498db, #2980b9);  // only for Opera
background-image: linear-gradient(to bottom, #3498db, #2980b9); // and this .. for the rest of the world..

and therewith can you fix it for specially browsers

and, you're welcome .. ;) gern geschehen :)
will nur ne gescheite cloud die man selbst betreiben kann und sich sicher sein kann,
daß die daten einem nicht fremd eingesehen werden, private is private and should remain/keep !...

best regards
Blacky

ChristophWurst:
@blackcrack I don't think those browser prefixes will help, now that the newest firefox even listens to -webkit prefixes. However, let me know if you can find a solution that works at least on Firefox AND Chrome, which are the browsers we target. Thanks

blackcrack:
well,
take a look ther ,
Chrome should use the webkit.. imho ..
http://stackoverflow.com/questions/5411026/list-of-css-vendor-prefixes

so +| -webkit-width: 230px ;   |

you should test it, because i have no chrome.. (dont like it.. FF only ;) )
best regards
Blacky

ChristophWurst:
Hi @blackcrack,
@skjnldsv fixed it for all browser: #1476, so I'm closing this one
Thanks!

blackcrack:
*thumps up ! *

well done !

best regards
Blacky

PR #1463, opened 2016-04-28
automatically adjust new message button width on resize and other rel…

### LINKED ISSUE #1357: New Mail Button overlays Scrollbar
## ownCloud Mail

The "New Mail" (Neue Nachricht in German) Button overlays the Scrollbar just a bit. The Window is maximised on a resoluten of 1920*1080, so there isn't any horizontal space Problem. However, I do have quite some accounts integrated, which explains the vertical scroll in the Folder View

### Steps to reproduce
1. Open Owncloud Mail (0.4) with **Firefox 45.0 on Windows 10**

### Expected behaviour
Button shouldn't overlay scrollbar

### Actual behaviour
it does

**OwnCloud Mail version:** (see ownCloud apps admin page)
0.4.0

## ownCloud Core
0.9

unbenannt

ChristophWurst:
…evant events

this replaces #1461
fixes #1357

@Maulwurf16 @jancborchardt @skjnldsv

### DIFF
--- a/js/views/navigation.js
+++ b/js/views/navigation.js
@@ -5,7 +5,7 @@
  * later. See the COPYING file.
  *
  * @author Christoph Wurst <christoph@winzerhof-wurst.at>
- * @copyright Christoph Wurst 2015
+ * @copyright Christoph Wurst 2015, 2016
  */
 
 define(function(require) {
@@ -32,9 +32,14 @@ define(function(require) {
 
 			this.listenTo(Radio.ui, 'navigation:show', this.show);
 			this.listenTo(Radio.ui, 'navigation:hide', this.hide);
+			this.listenTo(Radio.ui, 'navigation:resize', this._adjustButtonWidth);
+			this.listenTo(Radio.ui, 'window:resize', this._adjustButtonWidth);
 		},
 		render: function() {
 			// This view doesn't need rendering
+		},
+		onShow: function() {
+			this._adjustButtonWidth();
 		},
 		show: function() {
 			this.$el.show();
@@ -46,6 +51,11 @@ define(function(require) {
 				this.$el.hide();
 				$('#app-navigation-toggle').css('background-image', 'none');
 			}
+		},
+		_adjustButtonWidth: function() {
+			var newMailButtonContainer = $('#mail-new-message-fixed');
+			var innerWidth = $('#app-navigation-accounts').width();
+			newMailButtonContainer.attr('style', 'width: ' + innerWidth + 'px !important');
 		}
 	});
 });
--- a/js/views/account.js
+++ b/js/views/account.js
@@ -13,13 +13,14 @@ define(function(require) {
 
 	var Backbone = require('backbone');
 	var Handlebars = require('handlebars');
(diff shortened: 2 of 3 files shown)


js/views/navigation.js
```diff
@@ -46,6 +51,11 @@ define(function(require) {
 				this.$el.hide();
 				$('#app-navigation-toggle').css('background-image', 'none');
 			}
+		},
+		_adjustButtonWidth: function() {
+			var newMailButtonContainer = $('#mail-new-message-fixed');
+			var innerWidth = $('#app-navigation-accounts').width();
+			newMailButtonContainer.attr('style', 'width: ' + innerWidth + 'px !important');
```
ChristophWurst: This looks bad, but since we already make use of ``!important`` in the css file, we also have to enforce it here.

skjnldsv: The question is why is it on the css in the first place? :)

ChristophWurst: Navigation entries have 100% width. As the button has a fixed position, this would make it as wide as the browser window. Hence, the width is enforced on that specific element. I tried to remove it, but I couldn't get it to an acceptable state. If you know a fix for that, let me know ;-)

skjnldsv: Dammit, I shoud I kept my mouth shut! :p 
I will check on this later on the weekend!

ChristophWurst: @skjnldsv ping :)

skjnldsv: I have a 4day week-end tonight. Will be good :)

jancborchardt:
:D I mean honestly I would just cut the background of the button and be done with it. Especially since we collapse folders now, scrolling isn't that prevalent in the sidebar anymore.

Gomez:
For me this solution looks good. 

After rebase i would merge.

ChristophWurst:
@Gomez rebased

js/views/navigation.js
skjnldsv: Okay, I've given a thought. Move the first li (the button container) above the ul.
capture d ecran_2016-05-05_07-51-31

Remove the unused div container of the button (not the essential part, but clean code are always welcome :p )
And apply the following css (do it in console if you want, should override the header's css).

```
#mail-new-message-fixed {
    z-index: 1;
    padding-bottom: 12px;
    background-color: rgba(255, 255, 255, .75);
    padding: 10px;
}
#mail_new_message {
    width: 100%;
    padding: 10px;
    padding-left: 34px;
    background-position: 10px center;
    text-align: left;
}
.mail-account-email:first-child {
    display: none;
}
#app-navigation {
    overflow: hidden;
}
#app-navigation > ul {
    padding-bottom: 80px; /* scroll hack because of #settings fixed */
}
```

ChristophWurst: @skjnldsv could you please push that as a PR so we can check it out easily? :)

skjnldsv: Coming up! ;)

ChristophWurst:
replaced by a way better solution in #1476

PR #1464, opened 2016-04-29
return UID correctly so old drafts are automatically removed

### LINKED ISSUE #1450: Drafts: Lot of drafts created
Drafts seem to be created a lot and not purged. When a message is sent, all drafts associated with it should probably be deleted. Same as when a new draft of a message is created, the older draft should probably be overwritten. cc @irgendwie 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/33276673-drafts-lot-of-drafts-created?utm_campaign=plugin&utm_content=tracker%2F216487&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F216487&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ChristophWurst:
fixes #1450

@jancborchardt  please review

### DIFF
--- a/js/service/messageservice.js
+++ b/js/service/messageservice.js
@@ -290,10 +290,10 @@ define(function(require) {
 							collection.set([message], {remove: false});
 						}
 					}
-					defer.resolve(arguments);
+					defer.resolve(data);
 				},
 				error: function() {
-					defer.reject(arguments);
+					defer.reject();
 				},
 				data: {
 					to: message.to,
--- a/js/views/composer.js
+++ b/js/views/composer.js
@@ -126,6 +126,7 @@ define(function(require) {
 			if (!this.isReply()) {
 				this.accounts = options.accounts;
 				this.account = options.account || this.accounts.at(0);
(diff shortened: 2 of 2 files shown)


jancborchardt:
Might fix the multiple drafts problem while writing a new message. But editing a draft still keeps the old one and creates a new thing:
capture du 2016-05-02 17-16-52

ChristophWurst:
I guess you mean the following workflow:

1. start writing a new message, stop, wait so the draft is saved to the server
2. open the drafts folder
3. open the newly created draft
4. resume writing
5. another draft is saved instead of replacing the previous one

ChristophWurst:
In any case, there was an issue with that workflow as the UID of the original draft was not sent when updating it, hence it was not 'replaced'. Should be good now. Let me know if there are any other major issues with drafts, thanks.

jancborchardt:
Okidoke, thanks a lot! :+1:

PR #1465, opened 2016-05-01
remove unused require shims

ChristophWurst:
Those are left-overs from the first use of requirejs in this app. Apparently this was never really used, so we can savely remove it.

@tahaalibra please review

### DIFF
--- a/js/require_config.js
+++ b/js/require_config.js
@@ -7,7 +7,7 @@
  * later. See the COPYING file.
  *
  * @author Christoph Wurst <christoph@winzerhof-wurst.at>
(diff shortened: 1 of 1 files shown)


tahaalibra:
we may as well remove shim from js/tests/test-main.js...

ChristophWurst:
> we may as well remove shim from js/tests/test-main.js...

Yes, good catch! Could you please fix that and open a PR? :)

ChristophWurst:
(not merging this yes because it's scheduled for 0.4.5)

PR #1466, opened 2016-05-01
don't include js tests when packaging

ChristophWurst:
There is no reason to include those in the archive.

@tahaalibra please review


tahaalibra:
looks good...do we need to exclude readme files too?

ChristophWurst:
> do we need to exclude readme files too?

No, the archive should contain the readme

PR #1470, opened 2016-05-02
remove unused require shims from tests

tahaalibra:
Following #1465, also removing left-over shims of requirejs from the tests.


ChristophWurst:
:+1:

PR #1471, opened 2016-05-03
show first folder of an account incase of invalid folder id

tahaalibra:
if the folder id is not found then the user will be sent to the first folder of that account

please review @ChristophWurst

### DIFF
--- a/js/routecontroller.js
+++ b/js/routecontroller.js
@@ -108,9 +108,8 @@ define(function(require) {
 
 			var folder = account.get('folders').get(folderId);
 			if (_.isUndefined(folder)) {
-				// TODO: show first folder of this account
-				_this.default();
-				return;
+				folder = account.get('folders').at(0);
+				this._navigate('accounts/' + accountId + '/folders/' + folder.get('id'));
 			}
 			FolderController.showFolder(account, folder, noSelect);
 		},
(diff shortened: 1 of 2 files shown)


js/routecontroller.js
```diff
@@ -108,9 +108,8 @@ define(function(require) {
 
 			var folder = account.get('folders').get(folderId);
 			if (_.isUndefined(folder)) {
-				// TODO: show first folder of this account
-				_this.default();
-				return;
+				folder = account.get('folders').at(0);
```
ChristophWurst: Note: you could also use [``.first()``](http://backbonejs.org/#Collection-Underscore-Methods), but ``.at(0)`` is also fine as it does the same thing ;-)

ChristophWurst:
Changes look good, I'll test this later! Thanks a lot :-)

ChristophWurst:
@tahaalibra is there any case where this is currently used? (I don't really know how to test it)

tahaalibra:
actually i am not sure if there are in test cases...i tested it with the browser, by simply entering owncloud/apps/mail#accounts/1/folders/(placing some random text)

ChristophWurst:
- [ ] TODO: update changelog

tahaalibra:
i messed up on rebase..i will fix this now

ChristophWurst:
@tahaalibra please rebase on current master, there should only be one commit. Let me know if you need help :)

ChristophWurst:
And please adjust the changelog accordingly, this fix is part of 0.4.5 not 0.4.4 ;-)

ChristophWurst:
:+1: works perfectly! Thanks

PR #1477, opened 2016-05-05
update changelog for 0.4.4

ChristophWurst:
I updated the changelog following advices on http://keepachangelog.com/.

We should add a changelog entry for EVERY feature/fix/whatever

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,56 +1,71 @@
 # Changelog
+All notable changes to this project will be documented in this file.
 
-## 0.4.3
- * Improved autoconfig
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Btw. do we need to mention the developer of the PR? To be fair we should either also mention the one who discovered/reported the bug as this is often more work to do than fixing it, or don't mention anyone at all. Personally, I prefer the latter.

Gomez:
Ok, lets drop the name. 
The ### Added, ### Changed, ### Fixed is more work. Is it worth it? 

I will merge an we think about it until the next rls...

ChristophWurst:
> Is it worth it?

I think so, yes ;-)

Issue #1480, opened 2016-05-07
Update / Install Fail on Mysql 5.6


drfuture:
### Steps to reproduce
1. upgrade from 4.3
2. i think it''s the same on install.

### Expected behaviour
no problem

### Actual behaviour
Doctrine\DBAL\Exception\DriverException: An exception occurred while executing 'CREATE TABLE `oc_mail_collected_addresses` (`id` INT AUTO_INCREMENT NOT NULL, `user_id` VARCHAR(64) DEFAULT '' NOT NULL, `email` VARCHAR(256) NOT NULL, INDEX mail_collected_addr_userid_index (`user_id`), INDEX mail_collected_addr_email_index (`email`), PRIMARY KEY(`id`)) DEFAULT CHARACTER SET utf8 COLLATE utf8_bin ENGINE = InnoDB':

SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was too long; max key length is 767 bytes

### ownCloud Mail
0.4.4

### Information
i searched to find a solution, the proflem is that the varchar-size should be 255 max.
if it is 256 as on oc_mail_collected_addresses / email and Enginge is innoDB and Collate utf8 then the index could not created because the max size of 767byte (its mysqld internal) is to low.

The database.xml must be changed to field-size of 255.

ChristophWurst:
Hi @drfuture,
good catch! Since you already know how to fix that, it would be really cool if you could create a pull request ;-)

PR #1484, opened 2016-05-09
Narrow address column width to 255

drfuture:
Fix for Bug #1480


ChristophWurst:
@drfuture thanks a lot for fixing this so quickly and sending the PR!

drfuture:
No Problem :)

PR #1501, opened 2016-05-27
0.5.0 version bump

Gomez:
Prepare it.

@ChristophWurst Lates mater is good?


ChristophWurst:
> Lates mater is good?

I guess :speak_no_evil:

Issue #1503, opened 2016-05-28
Cannot Navigate to other Folders, Only Inbox accessible


joe-liberty:
### Steps to reproduce
1. Mail, go to another folder underneath inbox
2.
3.

### Expected behaviour
Tell us what should happen

It should open the folder and load messages

### Actual behaviour
Tell us what happens instead

Instead it says "checking messages" and only Inbox is refreshed

### ownCloud Mail

**ownCloud Mail version:** (see ownCloud apps admin page)

0.5.0

**Mailserver or service:** (e.g. Outlook, Yahoo, Gmail, Exchange,...)
 
Runbox

**Transport security - IMAP:** (None, SSL, TLS, STARTTLS) 

IMAP, SSL, TLS

**Transport security - SMTP:** (None, SSL, TLS, STARTTLS)

IMAP, SSL, TLS

**Number of accounts:**

1

**ownCloud Mail version build date:** (only if you are using a Nightly Build)

### ownCloud Core

9.0.0

#### Server configuration
**Operating system**:

Ubuntu 14.04 LTS

**Web server:**

Apache2

**Database:**

MariaDB

**PHP version:**

5.5

**ownCloud version:** (see ownCloud admin page)

9.0.1

**Updated from an older ownCloud or fresh install:**

**List of activated apps:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your ownCloud installation folder

Enabled:
  - activity: 2.2.1
  - calendar: 1.2.2
  - comments: 0.2
  - contacts: 1.2.0.0
  - conversations: 0.2.3
  - dav: 0.1.6
  - documents: 0.12.0
  - external: 1.2
  - federatedfilesharing: 0.1.0
  - federation: 0.0.4
  - files: 1.4.4
  - files_antivirus: 0.8.0.1
  - files_external: 0.5.2
  - files_mv: 0.8.2
  - files_pdfviewer: 0.8
  - files_sharing: 0.9.1
  - files_texteditor: 2.1
  - files_trashbin: 0.8.0
  - files_versions: 1.2.0
  - files_videoplayer: 0.9.8
  - firstrunwizard: 1.1
  - gallery: 14.5.0
  - mail: 0.5.0
  - music: 0.3.11
  - news: 7.1.2
  - notifications: 0.2.3
  - ownbackup: 0.3.8
  - provisioning_api: 0.4.1
  - registration: 0.1.3
  - systemtags: 0.2
  - templateeditor: 0.1
[...]

PR #1505, opened 2016-05-29
traverse folders correctly when searching by folder id

ChristophWurst:
The folder data structure is a tree, hence we need to traverse down all hierarchies when doing a search by id.

Note: one could optimize the algorithm by decoding the folder id and guessing the path. I'll leave that for the Mail 10.0 release :hear_no_evil: 

fixes https://github.com/owncloud/mail/issues/1503

@tahaalibra @Gomez @DeepDiver1975 @clementhk @Scheirle review please

### DIFF
--- a/js/models/account.js
+++ b/js/models/account.js
@@ -27,6 +27,41 @@ define(function(require) {
 		initialize: function() {
 			this.set('folders', new FolderCollection(this.get('folders')));
 		},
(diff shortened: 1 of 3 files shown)


Gomez:
Thanks! Works

joe-liberty:
Thanks for your help with this!

Issue #1513, opened 2016-06-02
Lots of errors while setting a an auto-configured account


ChristophWurst:
### Steps to reproduce
1. Set up account using the autoconfig method

### Expected behaviour
No errors. Connection errors should be caught and discarded.

### Actual behaviour
:boom: :boom: :boom: 

### ownCloud Mail

**ownCloud Mail version:**
0.5.1

##### Web server error log
```
[Thu Jun  2 21:17:18 2016] setting up auto detected account
[Thu Jun  2 21:17:18 2016] IsbDb: querying <winzerhof-wurst.at>
[Thu Jun  2 21:17:18 2016] IsbDb: querying <winzerhof-wurst.at> via <https://autoconfig.winzerhof-wurst.at/mail/config-v1.1.xml>
[Thu Jun  2 21:17:18 2016] file_get_contents(): php_network_getaddresses: getaddrinfo failed: Name or service not known at /home/christoph/workspace/owncloud/apps/mail/lib/service/autoconfig/ispdb.php#55
[Thu Jun  2 21:17:18 2016] file_get_contents(https://autoconfig.winzerhof-wurst.at/mail/config-v1.1.xml): failed to open stream: php_network_getaddresses: getaddrinfo failed: Name or service not known at /home/christoph/workspace/owncloud/apps/mail/lib/service/autoconfig/ispdb.php#55
[Thu Jun  2 21:17:18 2016] IsbDb: <https://autoconfig.winzerhof-wurst.at/mail/config-v1.1.xml> request timed out
[Thu Jun  2 21:17:18 2016] IsbDb: querying <winzerhof-wurst.at> via <https://winzerhof-wurst.at/.well-known/autoconfig/mail/config-v1.1.xml>
[Thu Jun  2 21:17:19 2016] file_get_contents(https://winzerhof-wurst.at/.well-known/autoconfig/mail/config-v1.1.xml): failed to open stream: HTTP request failed! HTTP/1.1 404 Not Found
 at /home/christoph/workspace/owncloud/apps/mail/lib/service/autoconfig/ispdb.php#55
[Thu Jun  2 21:17:19 2016] IsbDb: <https://winzerhof-wurst.at/.well-known/autoconfig/mail/config-v1.1.xml> request timed out
[Thu Jun  2 21:17:19 2016] IsbDb: querying <winzerhof-wurst.at> via <https://autoconfig.thunderbird.net/v1.1/winzerhof-wurst.at>
[...]

PR #1515, opened 2016-06-02
Ignore errors while querying autoconfig urls.

Scheirle:
This removes most of the errors reported in #1513, but not all.

### DIFF
--- a/lib/service/autoconfig/ispdb.php
+++ b/lib/service/autoconfig/ispdb.php
@@ -48,15 +48,18 @@ class IspDb {
 
 	private function queryUrl($url) {
 		try {
+			set_error_handler(function() { /* ignore all errors */ });
 			$content = @file_get_contents($url, false, stream_context_create([
 				'http' => [
 					'timeout' => 7
 				]
 			]));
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Wow. You're fast :o

ChristophWurst:
Hm, I still see many errors. Can't tell if there are fewer when using this patch :-/

Scheirle:
My mistake, wasn't in debug mode so the messages got suppressed.

I now pushed a version which removes the error handler and restores it afterwards.
I'm not sure if this is the best solution.

lib/service/autoconfig/ispdb.php
```diff
@@ -48,15 +48,18 @@ public function __construct(Logger $logger) {
 
 	private function queryUrl($url) {
 		try {
+			set_error_handler(function() { /* ignore all errors */ });
```
ChristophWurst: I don't have a good feeling adding that :-/

ChristophWurst:
I'm closing this PR due to inactivity
Thanks :-)

PR #1521, opened 2016-06-06
popover should close on clicking somewhere else

tahaalibra:
Fix: the popover should close when clicking somewhere else.

@Gomez @ChristophWurst

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 
 ## 0.5.2 - unreleased
 
+### Fixed
+- Close popover on clicking somewhere else
+  [#1521](https://github.com/owncloud/mail/pull/1521) @tahaalibra
+
 ## 0.5.1 – 2016-05-30
 
 ### Fixed
--- a/js/views/account.js
+++ b/js/views/account.js
@@ -65,14 +65,6 @@ define(function(require) {
 		toggleCollapse: function() {
 			this.collapsed = !this.collapsed;
 			this.render();
-			this.listenTo(Radio.ui, 'document:click', function(event) {
-				var target = $(event.target);
-				if (!this.$el.is(target.closest('.navigation-account'))) {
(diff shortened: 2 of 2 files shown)


js/views/account.js
```diff
@@ -65,6 +65,10 @@ define(function(require) {
 		toggleCollapse: function() {
 			this.collapsed = !this.collapsed;
 			this.render();
+		},
+		toggleMenu: function(e) {
+			e.preventDefault();
+			this.menuShown = !this.menuShown;
 			this.listenTo(Radio.ui, 'document:click', function(event) {
```
ChristophWurst: Hm, doesn't that mean everytime the menu is toggled, another event handler is registered?

ChristophWurst: add a ``onShow`` method and move that event handler registration in there. That method is called by Marionette once a view has finished rendering and is added to the DOM.

tahaalibra:
@ChristophWurst putting the listening in initialize...does this solve the problem

ChristophWurst:
> @ChristophWurst putting the listening in initialize function...does this solve the problem

Well, did you try the ``onShow`` approach? I'm in favor of that since the initialize method might be called even if the view is not attached to the DOM, which is error-prone…

tahaalibra:
@ChristophWurst using onShow , thanks for help !!...

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 
 ## 0.5.2 - unreleased
 
+### Fixed
+- Close popover on clicking somewhere else
```
ChristophWurst: awesome :-)

ChristophWurst:
:+1: very nice!

ChristophWurst:
oh, now there are 4 commits. Can you squash them with ``git rebase -i HEAD~4``? Thanks

ChristophWurst:
ah, you're fast :rocket:

PR #1526, opened 2016-06-12
change to secure connection

enoch85:
cc @ChristophWurst  @Gomez

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
 	"repositories": [
 		{
 			"type": "pear",
-			"url": "http://pear.horde.org"
+			"url": "https://pear.horde.org"
 		}
 	],
 	"require": {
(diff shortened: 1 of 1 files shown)


ChristophWurst:
maybe we can then remove the ``"secure-http": false`` option too

enoch85:
Done

ChristophWurst:
Thanks @enoch85!

ChristophWurst:
:+1: :rocket:

enoch85:
Was a bit too quick there: https://getcomposer.org/doc/06-config.md#secure-http

Think we need the config option.

enoch85:
No, this need to be reverted. I didn't check before, but it seems like we need to generate a SSL certificate to be able to get files over SSL. I thought it was just a simple change, but no.

PR #1527, opened 2016-06-12
Revert "change to secure connection"

enoch85:
Reverts owncloud/mail#1526

Merged to quick before testing properly.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,8 +1,11 @@
 {
+	"config": {
+		"secure-http": false
+	},
 	"repositories": [
 		{
 			"type": "pear",
-			"url": "https://pear.horde.org"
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Why do you want to revert it? [It worked](https://travis-ci.org/owncloud/mail/builds/137032195)?!

enoch85:
Did you test it?

ChristophWurst:
Yes, works as expected. I think we used http for pear in the past as they didn't serve their packages via https.

ChristophWurst:
Also CI would catch those errors as it downloads the packages too

enoch85:
Sorry, my misstake. It was a temporary error on my side. Everything is ok! :)

PR #1535, opened 2016-06-19
Update composerlock to https urls

### LINKED ISSUE #1534: Update of composer.lock to https only
On one of my prod server i got this error:

```
[Composer\Downloader\TransportException]                                                                                                                                     
  Your configuration does not allow connections to http://pear.horde.org/get/Horde_Exception-2.0.8.tgz. See https://getcomposer.org/doc/06-config.md#secure-http for details. 
```
Composer was set to https (#1526)only but the update for the lock file was missing.

Gomez:
fixes #1534

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -101,7 +101,7 @@
             "version": "1.1.1",
             "dist": {
                 "type": "file",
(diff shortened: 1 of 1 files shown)


ChristophWurst:
:+1:

Issue #1151, opened 2015-11-05
Line breaks inserted for sent mail [$50]


jancborchardt:
I just sent a mail out of the unified inbox and it seems like the Mail app forced some line breaks in there to keep it to ~70 characters a line. This should not happen. :)

@ChristophWurst @Gomez @irgendwie we didn’t do that before, right?

ChristophWurst:
was the original mail a HTML mail?

jancborchardt:
Nope, this happens when I write a completely new mail from the Mail app. Just did it (hacking here with @Gomez) – if you want to look into it then we will check out something else. :)

ChristophWurst:
Hm, strange. I hope you guys are hacking on the mail app ;-)
I'm too busy for mail app hacking at the moment, I just had a look at the last remaining issues for 0.3.

jancborchardt:
Interestingly enough, this doesn’t show in other clients such as Thunderbird. Does this have something to do with link detection or something else we parse in the text?

Gomez:
Ok, this happens here:
https://github.com/owncloud/mail/blob/master/lib/account.php#L177

Before `$raw` everything is fine, mail gets send without line breaks. Than `$mail->getRaw()` messed it up.
@ChristophWurst You have a idea?

ChristophWurst:
@Gomez the message is sent [here](https://github.com/owncloud/mail/blob/master/lib/account.php#L172). If ``$mail->getRaw()``then inserts those mysterious line breaks than that should only effect the message that is saved to the sent-folder, but not the actual message that is sent via SMTP.

I tried to locate the bug some weeks ago by going through the code step-by-step but couldn't really find out what's going wrong there.

jancborchardt:
Why should those line breaks even be in the message which is saved to the sent folder?

Gomez:
@ChristophWurst Right. The message gets send correctly. The linebreaks are only in the sent folder. Still Thunderbird dont show them. I will dig deeper...

Gomez:
Ok, this drives my crazy :( With saveDrafts its the same (easier to reproduce):
https://github.com/owncloud/mail/blob/master/lib/account.php#L219
Reproduce
1. Click "New message"
2. Fill in body text without line breaks 
3. Click on drafts folder 
= line breaks in the saved draft (<br>)

I looked into horde/imp how they save the send mails, but cant really find it. getRaw() is rarly used. I suspect there is something wrong with toString() or the stream (inside the horde stuff)?

[...]

Issue #782, opened 2015-06-17
Rich Text E-Mails with TinyMCE


stefan-niedermann:
It would be nice, if one had the option write Rich Text E-Mails via TinyMCE. At least a **bold** and *cursive* should be possible ;)

wurstchristoph:
Also needed for this feature:  #605

aspdye:
Why is this in Backlog? I think it is an "Basic Feature". Aren't we able to take the Editor from the Notes App?

Issue #1520, opened 2016-06-05
Rich text editor


ChristophWurst:
At some point we should add support for [HTML message](https://github.com/owncloud/mail/issues/782). As we will have to include some kind of rich javascript editor, I'd like to use this ticket for a comparison and for voting.

- [ ] Research which editors could be used and add one comment per editor
- [ ] Have a voting on which editor we want to choose

If anybody know good javascript editor, please add them to the list. Please also list pros and cons if possible.

ChristophWurst:
# SimpleMDE

todo

ChristophWurst:
# TinyMCE

todo

ChristophWurst:
# Trumbowyg

URL: https://github.com/Alex-D/Trumbowyg
DEMO: http://alex-d.github.io/Trumbowyg/

## Pro
* Clean, simple user interface
* Easy to use

## Con
* Demo does not work on Firefox, but on Chrome?

ChristophWurst:
# TinyEditor

URL: https://github.com/jessegreathouse/TinyEditor
DEMO: http://sandbox.scriptiny.com/tinyeditor/

## Pro

## Con
* Looks ugly :see_no_evil:

Gomez:
Need to take a closer look:
https://prosemirror.net/

ChristophWurst:
# Promisemirror

URL: https://github.com/prosemirror/prosemirror
DEMO: https://prosemirror.net/

## Pro
* Simple, clean interface

## Con

jancborchardt:
Maybe go for TinyMCE cause it’s the most popular one?

xlyz:
ckeditor?

jancborchardt:
CKEditor has a quite outdated interface look and a whole bunch of stuff in the toolbar. I’d say we not go for that.

xlyz:
are you aware that ckeditor is skinnable and you can select which functionality include (and/or enable/disable)?
http://ckeditor.com/demo
http://ckeditor.com/builder

Henni:
:+1: for TinyMCE, widely used and tested with Wordpress.

Alternatively ProseMirror, but seems like it is not completely mature yet.

Issue #605, opened 2015-05-05
Support of HTML drafts


wurstchristoph:
Some mail clients create HTML drafts which cannot be opened at the moment, a error is shown instead.

For integration with other mail clients also HTML drafts have to be supported. This may need some advanced (HTML) editor to work properly.

PR #1536, opened 2016-06-20
added html email support

haseebeqx:
added html email composing support


js/util/trumbowyg/hyperlink.js
```diff
@@ -0,0 +1,87 @@
+(function ($) {
+	'use strict';
+	$.extend(true, $.trumbowyg, {
+	langs: {
```
ChristophWurst: please install js dependencies via bower

ChristophWurst:
Hi @haseebeqx,
thanks a lot for working on this! This is super awesome! Why did you choose trumbowyg? I recently opened an issue to discuss which rich text editor we should go for and AFAIK we agreed on TinyMCE. See https://github.com/owncloud/mail/issues/1520

@Gomez @jancborchardt do we want to add html editor support now? You guys voted against it one week ago.

js/util/trumbowyg/hyperlink.js
haseebeqx: that is a utility i created for making hyperlink compatible with email. default hyperlink in the dependency contained target and title so i had to extend it

haseebeqx:
ok. i was unaware of the discussion. sorry about that. 
i thought trumbowyg is enough for the need as an email composer.

ChristophWurst:
Hm, this looks weird on FF and Chrome
bildschirmfoto von 2016-06-20 13-32-37

haseebeqx:
what! this could  be a problem with my ide on handling git commit. let me check it
screenshot from 2016-06-20 20-06-34

haseebeqx:
fixed now

ChristophWurst:
@haseebeqx is there any chance that can add an option (e.g. checkbox) to switch to text-only messages? Personally, I prefer to not use HTML messages if possible.

haseebeqx:
@ChristophWurst  ok i will add that option. current implementation is compatible with that

ChristophWurst:
we also have to check whether this is compatible with (resuming) drafts

ChristophWurst:
- [ ] TODO: resuming drafts does not work as the HTML is escaped when loaded into the editor (one sees all the HTML tags that were created)

ChristophWurst:
- [ ] TODO: the editor and the inputs above it should stick together. currently there is space between them

ChristophWurst:
Interesting, I don't see the link menu that you have on your screenshot

haseebeqx:
@ChristophWurst updated with toggle button

haseebeqx:
now!. how can i resume drafts at all. i was unaware of that :/  but may be fixed in last update

ChristophWurst:
@haseebeqx open your drafts folder and select a draft ;-)

ChristophWurst:
- [ ] TODO: check whether this is compatible with the Mailvelope browser extension

ChristophWurst:
Sorry @haseebeqx for adding more and more items to the todo list, but if we add a html editor we want do do it right and not break existing features. Thanks a lot for helping with this!

haseebeqx:
@ChristophWurst  ok this will fix drafts :)

haseebeqx:
@ChristophWurst i tested mailvelope plugin it looks compatible with it. is there any bugs ?.

Gomez:
The editor looks good and simple. 

Maybe we should make text the default and switch to html with this button? @jancborchardt @ChristophWurst Agree? Later on we can add a config.php to switch this?

This looks brocken, the top border is gone:

area_bug

Gomez:
Some mails look in the reply like this:

garbage

Gomez:
But great work @haseebeqx! Looking forward to get this merged!

haseebeqx:
> Some mails look in the reply like this
@Gomez 
which is the line ending. may be something other than \n.

haseebeqx:
you may have to clear cache to make this fix work on browser

haseebeqx:
@Gomez if this was a problem with line endings this update will fix. can you confirm?

ChristophWurst:
- [ ] TODO: some line endings are messed up for replies and the quoted text is one single line

js/views/composer.js
```diff
@@ -187,6 +208,22 @@ define(function(require) {
 				_this.saveDraft();
 			}, this.draftInterval);
 		},
+		toggleEditor: function()
```
ChristophWurst: Hmmm… I thought that we could use the featured editor, but disable all html-like features. Also I'd like to have the checkbox in the editor's tools menu. Would that be possible with trumbowyg?

js/views/composer.js
```diff
@@ -218,13 +255,18 @@ define(function(require) {
...
 			message.bcc = bcc.val();
 			message.subject = subject.val();
 			message.attachments = this.attachments.toJSON();
-
+			if($('.toggle-editor').prop('checked')) {
+				message.type = 'text/plain';
+				message.body = newMessageBody.val();
+			}else{
+				message.type = "text/html"
+				message.body = newMessageBody.trumbowyg('html').replace('<br>&gt;','\n>');
```
ChristophWurst: what are you doing here? I guess if the content is HTML we shouldn't care about '\n's, should we?

js/views/composer.js
```diff
@@ -393,14 +439,17 @@ define(function(require) {
...
+				$.datepicker.formatDate('D, d. MM yy ', date) +
+				date.getHours() + ':' + (minutes < 10 ? '0' : '') + minutes + '<br>&gt; ' +
+				text.replace(/(?:\r\n|\r|\n)/g, '<br>&gt;');
+			this.replyText = 	'\n\n\n' +
 				from + ' – ' +
 				$.datepicker.formatDate('D, d. MM yy ', date) +
 				date.getHours() + ':' + (minutes < 10 ? '0' : '') + minutes + '\n> ' +
-				text.replace(/\n/g, '\n> ')
-				);
+				text.replace(/(?:\r\n|\r|\n)/g, '\n> ');
```
ChristophWurst: please elaborate on what this does

lib/model/imapmessage.php
```diff
@@ -479,6 +479,9 @@ public function getFullMessage($ownMail, $specialRole=null) {
 		$data = $this->getListArray();
 		if ($this->hasHtmlMessage) {
 			$data['hasHtmlBody'] = true;
+			if($specialRole === 'drafts'){
+				$data['body'] = rtrim($this->htmlMessage,"<br><br>");
```
ChristophWurst: what are you trimming here?

ChristophWurst:
apparently you changed file permissions on some files ``lib/model/message.php 100644 → 100755``. Was that intended?

ChristophWurst:
@Gomez @jancborchardt @Henni before too much effort is put into this, is trumbowyg the right choice? We did not agree on anything in https://github.com/owncloud/mail/issues/1520. One important criteria for selecting the editor is the ability to add a plugin that allows to write flowed text to fix https://github.com/owncloud/mail/issues/1151#issuecomment-218396979

ChristophWurst:
> @ChristophWurst Agree? Later on we can add a config.php to switch this?

Nope, anything in config.php is a system setting. If we want to make it a setting then it's definitely a user-specific setting.

js/views/composer.js
haseebeqx: if editor converts \n> to <br>&gt; then it is replaced back \n >

js/views/composer.js
haseebeqx: handling multiple line endings

lib/model/imapmessage.php
haseebeqx: <br><br> is appended with htmlMessage by default i was removing that

js/views/composer.js
ChristophWurst: I don't get it, sorry. When is this conversion necessary?

js/views/composer.js
haseebeqx: trumbowyg only has viewhtml functionality. but we can add a plugin for that  if we want, in that case, any example links on how it should look like

js/views/composer.js
ChristophWurst: Isn't that a general problem with special html chars? try 'öäüê'

lib/model/imapmessage.php
ChristophWurst: so you're fixing the symptoms here. Can we fix the root cause instead? :-)

ChristophWurst: (not that github hides/replace some html tags like ``<br>``, make sure you escape those)

js/views/composer.js
ChristophWurst: Maybe like this
f113f23e-37d7-11e6-8dc8-3de13731873f
cc @skjnldsv @jancborchardt @Henni

ChristophWurst: if 'text only' mode is selected, the other tools should be either greyed out or hidden

js/views/composer.js
haseebeqx: fixed my previous comment

haseebeqx:
>Was that intended?

nope, that was not intented. may be accidental. should i change it back

ChristophWurst:
> should i change it back

yes, please

lib/model/imapmessage.php
haseebeqx: >Can we fix the root cause instead? :-)

what is a root cause?.

haseebeqx:
>yes, please

is that the only file?

lib/model/imapmessage.php
ChristophWurst: the root of this problem. We should debug where those line breaks come from and remove those there instead of patching this here. Does that make sense?

haseebeqx: ok

ChristophWurst:
> is that the only file?

have a look at the (github) diff and you'll see the other ones too, there are more files affected

lib/model/imapmessage.php
haseebeqx: the root cause is `handleHtmlMessage()` function at line 571. i think `&lt;br&gt&lt;br&gt;` appending is needed. what is your opinion

haseebeqx:
>TODO: some line endings are messed up for replies and the quoted text is one single line

@ChristophWurst that was not a problem with line ending. it was with `setReplyBody` not being called in `message.js`. it is fixed now.

js/views/composer.js
haseebeqx: should the content type change to 'text/plain' in that case

skjnldsv:
Looking fine here!! Nice job! :tada:

js/views/composer.js
haseebeqx: @ChristophWurst another doubt.  should the current formatting is removed when moving to text only

haseebeqx: Before clicking
htmlmode

after clicking
textonly

Gomez:
@haseebeqx The toggleeditor.js is missing.

mar1u5:
Uhmm... Nice work 👍 ...but can we offer more stuff (font size, pictures, ...) 😁

haseebeqx:
@mar1u5 loading an external image in editor causes content-securiy policy error

templates/index.php
```diff
@@ -24,6 +24,7 @@
  *
  */
 style('mail','../js/vendor/trumbowyg/dist/ui/trumbowyg.min');
+style('mail','../js/vendor/trumbowyg/dist/plugins/colors/ui/trumbowyg.colors.min');
```
ChristophWurst: what is this colors thing about?

js/util/trumbowyg/toggleeditor.js
```diff
@@ -0,0 +1,37 @@
...
+                  document.activeElement.blur();
+                  t.$box.toggleClass(prefix + 'editor-hidden ' + prefix + 'editor-visible');
+                  t.$btnPane.toggleClass(prefix + 'disable');
+                  $('.' + prefix + 'toggleeditor-button', t.$btnPane).toggleClass(prefix + 'active');
+                  if (t.$box.hasClass(prefix + 'editor-visible')) {
+                      t.$ta.attr('tabindex', -1);
+                  } else {
+                      t.$ta.removeAttr('tabindex');
+                  }
+              }, 0);
```
ChristophWurst: what's the 0 timeout needed for?

jancborchardt:
Just some design things:
- The default should remain plain text, with no controls. This is the most simple way to write email.
- The toggle should be labeled »Switch to rich-text editing« and placed on the right next to the »Add attachment from files« button.
- Even when replying to a HTML email, the controls should only show once the switcher is enabled.

Remember folks, let’s not go overboard with this. The main reason we’re introducing this is so reading, replying to, and forwarding of HTML mails is nicer, _not_ to encourage people to write with a whole bunch of styles.

jancborchardt:
By the way, please see for reference how Gmail / Google Inbox does it very nicely in an integrated way with just one styling button by default instead of a full bar.

ChristophWurst:
I had a look at the different [text editors](https://github.com/owncloud/mail/issues/1520) and I'm okay with using trumbowyg if we can style it to fit the app's design (e.g. adjust borders and background color of the tools menu).

ChristophWurst:
> The default should remain plain text, with no controls. This is the most simple way to write email.

I that's easily possible with trumbowyg – great, let's do that. If not greying out the tools is also fine by me.

js/views/composer.js
ChristophWurst: @Gomez and I tested your latest commit and we couldn't see that button on the interface.

jancborchardt:
If the tools are not used, they should not only be greyed out but not shown at all. ;)

templates/index.php
haseebeqx: text color and bg-color

js/util/trumbowyg/toggleeditor.js
haseebeqx: this plugin no longer needed. since graying out is rejected by @jancborchardt

haseebeqx:
@ChristophWurst @jancborchardt @Gomez i had a look at inbox's editor. looks good. html editor is not disabled by default. only difference is design bar is at the bottom.  should we use a similar design?. or  only a send/reply button and there will be a switch to rich text editor  near send/reply. which brings full editor
capture

haseebeqx:
>if we can style it to fit the app's design

that will be easy. it is just a css file

js/views/composer.js
haseebeqx: >we couldn't see that button on the interface.

may be a problem with cached js. did you tried opening in incognito mode to test cached js

Gomez:
>  The main reason we’re introducing this is so reading, replying to, and forwarding of HTML mails is nicer,
"reading" and "replying to" inside the text editor does not improve right now. Is it doable to show HTML mails correctly rendered inside trumbowyg? @haseebeqx

js/views/composer.js
Gomez: Does not help. Lets finish the discussion on how to implement and come back later to the details

jancborchardt:
> or only a send/reply button and a switch to rich text editor near send/reply

I’d prefer that for now to keep the visual noise to a minimum. We still need to polish other areas, so push the HTML editor too hard will make people think it’s too complete than it is.

haseebeqx:
@ChristophWurst @jancborchardt @Gomez now default is text-only mode. there is a toggle button to switch. we can push html editor to default when we are ready for it .

haseebeqx:
this fixes issues #782 and #605 hope this will get merged soon.

haseebeqx:
>what are your plans after you've finished this PR? Will you help maintain this and other features?

that will be awesome if that is an option

haseebeqx:
>I'd recommend on pausing development of this feature until we have come to an conclusion whether we (the mail app developers team) want that feature now and how that should look like

since there is a bounty backed by official owncloud on bountysource. i thought this is a must implement feature and just waiting for somebody to get implemented.

ChristophWurst:
Feel free to continue working on this. I just want to save you from wasting too much time on this in case we agree that we're not yet ready for a html editor. @jancborchardt and @Gomez were against that feature only two weeks ago, so I wonder if there opinion changed so quickly. And for my part, I'd like to have it, but only if we plan, design and implement it appropriately. The way it is now feels rushed and I'm afraid this will produce more problems than it actually solves. In that case I'd rather not have that feature and wait until we've found a solution that fits better. We (the ownCloud Community) have put the bounty on those important features especially because it's not trivial to implement. This does not necessarily mean we want those features immediately.
So please don't get me wrong. I value what you've done so far, but I can't say for sure whether we are ready for that feature. I've been the main developer of this app in the last months, hence I might also be the one that *has to* fix bugs that will come with this cool (but potentially buggy) feature.

Gomez:
I agree with @ChristophWurst, we need more time on this. 
This is a great base and we learn on the way what we need and what we don't need. 

Best would be pausing here, if you like you cloud go on with this: https://github.com/owncloud/mail/issues/363

jancborchardt:
Oh and when we should ever go forward with this: Interface-wise this unobtrusive bar which appears when marking text would be great: https://yabwe.github.io/medium-editor/demo.html (Also used by pretty popular blogging platform Medium)

Issue #1314, opened 2016-02-22
Backbone models


ChristophWurst:
Right now we're using Backbone models only for rendering Marionette collection views. By consistently using models throughout the client, we could lower the complexity of working with related data. 

I had a look at the JavaScript code and identified the following models:
- Account
- Folder
- Messge
- MessageFlag

![backbone_models](https://cloud.githubusercontent.com/assets/1374172/13213610/72a39a26-d94a-11e5-9329-931ab29f5458.png)

Using Backbone models has some advantages:
- By using bidirectional association between models it will be easier to work with related data
- Global state variables (e.g. inside the ``state`` module) can be eliminated because all the data is passed when calling a module function or triggering an event.
- We can use [Backbone sync](http://backbonejs.org/#Sync) on those models
  - Might be helpful to use [horde sync](https://github.com/owncloud/mail/pull/763) for folders
- Existing code will be easier to maintain
- It will be easier to implement new features

@owncloud/mail

ChristophWurst:
@DeepDiver1975 you also mentioned to fully migrate to backbone [once](https://github.com/owncloud/mail/issues/840#issuecomment-118425690) ;-)

jancborchardt:
We should also keep threading/conversations #21 in mind. Not sure if a conversation warrants an entire model, but definitely message should have some kind of »thread« parameter so we can group messages.

(We would only order messages flat and chronologically, so no need for overcomplicated parent/child relations.)

ChristophWurst:
> We should also keep threading/conversations #21 in mind.

Good point. I haven't thought about it thoroughly yet but it's likely to be some kind of reference from one message to the previous one in the thread.

jancborchardt:
Intermittent messages can be deleted though, so there maybe should be smth like a thread ID. Doesn't the email header have smth like that? I remember something similar.

DeepDiver1975:
We will add threading to the model once we know how to implement it.
Wild guessing will not help.

This model is good for the current application concept :+1:

jancborchardt:
Well, I just want to make sure that it’s futureproof. If it can be added at any point – fine.

ChristophWurst:
What's left:
- [ ] use a Message model for sending messages and drafts
- [ ] use a Backbone model for the message content
- [ ] add bidirectional navigation to models (e.g. save 'account' property on folders) to allow easier navigation and to reduce the number of parameters needed for methods/events

PR #1538, opened 2016-06-20
Message Backbone model/collection

ChristophWurst:
Messages are the only entity we do not yet manage as backbone model/collection. This is the first shot into fixing this.

The next steps will be to use backbone/horde sync to update folders.

ref #1314 

@Gomez @DeepDiver1975 @tahaalibra @irgendwie review please


Gomez:
Error on deleting mails:
callstack

ChristophWurst:
@Gomez fixed. Please test again

Gomez:
I took over to just fix the jscs errors. Hope you dont mind :)

ChristophWurst:
@Gomez rebased and fixed a small jasmine bug. Please re-review, I'd like to continue with this little cleanup :rocket:

Gomez:
backbone.radio was used before, how does this work?

ChristophWurst:
> backbone.radio was used before, how does this worked?

correct, but the dependency graph did not require to load it before. It's now a sub-dependency of a module that I touched with this PR

PR #1539, opened 2016-06-20
load more messages button is obsolete with infinite scroll

ChristophWurst:
A small UI enhancement, we don't need that button anymore.

bildschirmfoto von 2016-06-20 19-16-06

@jancborchardt @skjnldsv check it out!


skjnldsv:
:+1:

jancborchardt:
Works nicely! :+1:

PR #1543, opened 2016-06-28
[WIP] Pull to refresh

ChristophWurst:
https://github.com/apeatling/web-pull-to-refresh/

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -19,6 +19,7 @@
     "backbone.marionette": "^2.4.4",
     "backbone.radio": "^1.0.4",
     "domReady": "^2.0.1",
+    "hammerjs": "hammer.js#^2.0.4",
     "handlebars": "^4.0.5",
     "jquery": "2.2.3",
     "jquery-storage-api": "julien-maurel/jQuery-Storage-API#~1.7.4",
@@ -28,6 +29,7 @@
     "underscore": "~1.8.3",
     "davclient.js": "https://github.com/evert/davclient.js.git",
     "es6-promise": "^3.2.1",
-    "ical.js": "^1.2.1"
+    "ical.js": "^1.2.1",
+    "web-pull-to-refresh": "git@github.com:ChristophWurst/web-pull-to-refresh.git#^1.1.0"
   }
 }
--- a/css/mail.css
+++ b/css/mail.css
@@ -259,7 +259,6 @@
 	overflow-x: hidden;
 	overflow-y: auto;
 }
-#load-new-mail-messages,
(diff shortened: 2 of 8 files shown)


jancborchardt:
Nice! :)

In the [demo](https://apeatling.com/demos/web-ptr/) of that library you can only pull via actually clicking and pulling. On desktop of course it should be possible by simply scrolling.

bower.json
```diff
@@ -28,6 +29,7 @@
     "underscore": "~1.8.3",
     "davclient.js": "https://github.com/evert/davclient.js.git",
     "es6-promise": "^3.2.1",
-    "ical.js": "^1.2.1"
+    "ical.js": "^1.2.1",
+    "web-pull-to-refresh": "git@github.com:ChristophWurst/web-pull-to-refresh.git#^1.1.0"
   }
```
Gomez: Why is it forked and not used directly?

ChristophWurst: because the author could remove/rename the repo at any time and that shouldn't break our app

mar1u5: Would this also happen if we put this in as a GitModule :grin: (Dont know if this would make sense or wether this is possible) ...

ChristophWurst: put what in a git module?

mar1u5: `web-pull-to-refresh` :grin:

ChristophWurst: why should we if we can use bower to manage dependencies? That wouldn't solve any problem but make the install process for devs more complicated

ChristophWurst:
@jancborchardt if we remove the 'Check message' button on top of the messages list, how or where can we should loading feedback? Any idea? I had a quick look at other mail solutions with similar design (Kerio, Outlook online) and to my surprise, they do not have any loading feedback.
On Android the pull to refresh (PTR) of most apps shows a little spinner above (z-axis) the fist element of the list. I doubt that would fit the ownCloud design…

jancborchardt:
> On Android the pull to refresh (PTR) of most apps shows a little spinner above (z-axis) the fist element of the list.

Why would that not fit the design? :) When you pulled down the list, then that's exactly where the feedback should go. Of course the automatic refresh doesn't need feedback in this case.

And yeah, still very important for desktop is that the pull-to-refresh not only works by click-and-drag, but also by simply scrolling up.

jancborchardt:
cc @BernhardPosselt on this one because in the past we experimented with this in the News app and should coordinate there.

ChristophWurst:
JavaScript PTR libs are :hankey:

PR #1548, opened 2016-07-05
open account menu on right-click

ChristophWurst:
Small UX enhancement for people that are used to use right click for context menus.

@jancborchardt @skjnldsv what do you think?

### DIFF
--- a/js/views/account.js
+++ b/js/views/account.js
@@ -40,7 +40,8 @@ define(function(require) {
 		events: {
 			'click .account-toggle-collapse': 'toggleCollapse',
 			'click .app-navigation-entry-utils-menu-button button': 'toggleMenu',
-			'click @ui.deleteButton': 'onDelete'
+			'click @ui.deleteButton': 'onDelete',
+			'contextmenu': 'onRightClick'
 		},
 		ui: {
 			'menu': 'div.app-navigation-entry-menu',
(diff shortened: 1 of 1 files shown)


jancborchardt:
I can't decide between:
1. wow that's actually pretty cool!
2. hmm, people will expect this to work everywhere then for the action menu … I'd rather not start doing it.

So how about we also put it in other places like the files list to open the actions menu? And the news app, and Calendar app to open that menu. Would that work as nicely? I'd say it should be consistent, or nowhere. ;)

ChristophWurst:
Hm, good point. I'm afraid this cannot be done in a way that it automatically works for all apps by adding some JS magic to core. Hence either all apps have to implement that separately.

jancborchardt:
(If not, I’m fine with leaving it. But as said, we should either be consistent here or just not have it. :)

MorrisJobke:
> So, can we get that done?

I think so :) But having it in a way that it automatically works everywhere is quite hard I think. We need to make some intense testing.

jancborchardt:
Well I'm not too insistent that we need to have it. At the moment doing it right takes time from more important issues, and hence I'd say we leave it for now.

@ChristophWurst what do you think?

ChristophWurst:
Sure, let's leave that for now as that has low priority

PR #1549, opened 2016-07-05
open first mailbox when clicking the account's email address

ChristophWurst:
While switching between different account's inboxes, I found myself clicking on the account name instead of the 'inbox' mailbox. Hence, I thought we could load the first mailbox of the account in that situation, which most likely is the inbox of that account.

@jancborchardt @skjnldsv what do you think?

Not sure if we should also change the cursor accordingly.

### DIFF
--- a/js/views/account.js
+++ b/js/views/account.js
@@ -102,7 +102,9 @@ define(function(require) {
 		onClick: function(e) {
 			e.preventDefault();
 			if (this.model.get('folders').length > 0) {
-				Radio.navigation.trigger('folder', this.model.get('accountId'), this.model.get('folders').first().get('id'));
+				var accountId = this.model.get('accountId');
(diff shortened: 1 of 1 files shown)


jancborchardt:
Makes perfect sense indeed. :)

Trying this out with 2 accounts, it doesn't seem to work though. The account names / email addresses are not clickable or nothing happens.

ChristophWurst:
Tested again with single and two accounts: works as expected on FF and Chromium. The account email address is clickable (although the cursor does not change) and the first mailbox is loaded. If the inbox is selected already, the mailbox is reloaded.
Please have a look at your browser's debugging console. Do you see any logs/errors?

Gomez:
Works for me. :+1:

jancborchardt:
Ahhh … guess I forgot to run make on this one. ;)

PR #1550, opened 2016-07-07
smoother transition when adding another account

ChristophWurst:
this hides the message/folder content while loading the new account and shows a nice loading spinner instead.


Gomez:
:+1:

PR #1555, opened 2016-07-07
use 'mail' instead of 'Mail' for bower to prevent warning

ChristophWurst:
Fixes ```bower                     invalid-meta The "name" is recommended to be lowercase, can contain digits, dots, dashes```


Gomez:
Yes, warning gone. :+1:

Issue #160, opened 2014-08-06
Keyboard shortcuts


jancborchardt:
An essential feature for any good email client are good keyboard shortcuts to be able to quickly handle email:
* [ ] switch between open emails using `j` and `k` (like the News app) as well as cursor up & down
* [x] delete current email using `Del`
* [ ] star/unstar current email using `s`
* [ ] toggle read unread `m` (as noted by @zinks-)
* [ ] focus reply field with `r`
* other shortcuts?

(Of course they need to be disabled when you are in an input or textarea.)

jospoortvliet:
If up and down switch between mails, how will you scroll in a long email? I would suggest that the left/right keys should switch between mails, while up and down scroll within one email.
(I know a mail client that does that :D)

Other shortcuts: reply-to-all with 'a' and reply-to-list with l.

PR #1557, opened 2016-07-08
Enhanced keyboard event handling + new shortcuts

ChristophWurst:
This enhances/replaces the previous error-prone keyboard event handling and adds support for 'j', 'k', arrow left/right navigation of messages.

Note that I chose left/right error intentionally as up/down confuses FF and should be used to scroll in long messages or the messages list.

partially fixes https://github.com/owncloud/mail/issues/160

### DIFF
--- a/js/radio.js
+++ b/js/radio.js
@@ -11,27 +11,25 @@
 define(function(require) {
 	'use strict';
 
+	var _ = require('underscore');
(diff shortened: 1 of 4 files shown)


Gomez:
Nice, works correct. :+1:

jancborchardt:
Yeah, this is super awesome! :) :+1:

PR #1570, opened 2016-07-21
restore setup view state only on error

### LINKED ISSUE #1565: TypeError: _this.ui.iconLoading.hide is not a function
Happens while adding a (second) account:
bildschirmfoto von 2016-07-17 16-37-13

ChristophWurst:
if the setup was successful, we replace the setup view with a loading view.

fixes #1565

### DIFF
--- a/js/views/setup.js
+++ b/js/views/setup.js
@@ -160,13 +160,10 @@ define(function(require) {
 				});
 			});
 
-			$.when(creatingAccount).fail(function(error) {
-				Radio.ui.trigger('error:show', error);
-			});
-
 			var _this = this;
-			$.when(creatingAccount).always(function() {
+			$.when(creatingAccount).fail(function(error) {
 				_this.loading = false;
+				Radio.ui.trigger('error:show', error);
 				_this.ui.iconLoading.hide();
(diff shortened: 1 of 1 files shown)


ChristophWurst:
blocked CI-wise by https://github.com/owncloud/mail/pull/1571

ChristophWurst:
@jancborchardt test this please

ChristophWurst:
@owncloud/mail I need some review power on this :-)

jancborchardt:
Do we have some @owncloud/mail folks who want to test this as well? :) Come join the fun, Mail app development is at a fast pace!

jancborchardt:
@haseebeqx @Scheirle you up for testing this? :)

tahaalibra:
Great, Everything works fine. Just 1 thing i want to confirm that is on adding an account B, when a previous account A is already there, then on success the loading view loads account A. After adding a new account should we load the "All Inboxes" or the New Account that is added

ChristophWurst:
@tahaalibra good point. We shoud definitely either load the unified inbox or the newly added account

jancborchardt:
Yeah, good point – I’d say we should load the unified inbox. :)

PR #1573, opened 2016-07-22
increase min core version to 9.0 and remove all version hacks

### LINKED ISSUE #1572: Drop support for old ownCloud core versions
At the moment, we support 6 major versions of ownCloud. See https://github.com/owncloud/mail/blob/1b3c2438a85ea87cdb1efdba550b9c095b8dad87/appinfo/info.xml#L13

This means that in theory we have to manually test PRs and releases for several ownCloud versions. In practice, this does not happen. I'm developing on core master and occasionally I test stable releases on my production system (mostly the newest stable release of oc). Additionally the list of beta testers is really short, hence I don't feel comfortable specifying that this app works on older releases too.

From a technical point of view we could move forward faster and we'd have to make fewer (or maybe no) version hacks anymore and use the newer features/APIs oc/nc provides.

Therefore I would like to propose that we stop supporting older releases as long as this app in development. How about targeting 9.1 and master only?

cc @Gomez @DeepDiver1975 @jancborchardt @irgendwie @tahaalibra

ChristophWurst:
fixes #1572

TODO:
- [x] manual testing

### DIFF
--- a/lib/controller/pagecontroller.php
+++ b/lib/controller/pagecontroller.php
@@ -77,10 +77,8 @@ class PageController extends Controller {
 	 * @return TemplateResponse renders the index page
 	 */
 	public function index() {
-		// TODO: remove DEBUG constant check once minimum oc
-		// core version >= 8.2, see https://github.com/owncloud/core/pull/18510
 		$response = new TemplateResponse($this->appName, 'index', [
-			'debug' => (defined('DEBUG') && DEBUG) || $this->config->getSystemValue('debug', false),
+			'debug' => $this->config->getSystemValue('debug', false),
 			'app-version' => $this->config->getAppValue('mail', 'installed_version'),
 		]);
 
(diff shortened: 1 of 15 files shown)


lib/controller/pagecontroller.php
```diff
@@ -77,14 +77,11 @@ public function __construct($appName, IRequest $request, MailAccountMapper $mail
 	 * @return TemplateResponse renders the index page
 	 */
 	public function index() {
-		$coreVersion = $this->config->getSystemValue('version', '0.0.0');
-		$hasDavSupport = (int) version_compare($coreVersion, '9.0.0', '>=');
 		// TODO: remove DEBUG constant check once minimum oc
 		// core version >= 8.2, see https://github.com/owncloud/core/pull/18510
```
ChristophWurst: - [ ] TODO: remove

ChristophWurst:
rescheduled for 0.6 as this potentially breaks stuff.

jancborchardt:
Just remembered that if we do this we can also switch from `#mail-messages` to the standard `#app-content-list` from apps.css, as well as to `#app-content-detail`. Just like in Contacts.

ChristophWurst:
Did some manual testing and could not find any issues so far

jancborchardt:
Works! :+1:

PR #1579, opened 2016-07-28
remove 'Check messages' button and add scroll-to-refresh and shortcut

ChristophWurst:
TODO:
- [x] loading animation should be fixed (also shown when list is scrolled down)
- [x] weird bug causes a pagination request when scrolling down

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -259,7 +259,11 @@
 	overflow-x: hidden;
 	overflow-y: auto;
 }
-#load-new-mail-messages,
+#mail-message-list-loading {
+	position: absolute;
+	width: 100%;
+	height: 40px;
+}
 #load-more-mail-messages {
 	display: none;
 	margin: 10px auto;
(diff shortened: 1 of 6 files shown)


ChristophWurst:
bildschirmfoto von 2016-07-28 15-40-46
@skjnldsv could you help with the CSS please? The spinner is not 100% centered, I guess we need some flexbox magic for the message-list and message content, so we can position that spinner absolute. Do you have an idea how to do that? :-)

ChristophWurst:
:+1: for jan's part

jancborchardt:
:+1: for the whole thing :) very nice and finally that button is gone!

jancborchardt:
@skjnldsv if you want to play around with some flexbox stuff, feel free. :)

jancborchardt:
And cc @BernhardPosselt @cosenal on this. ;) For the Mail app we’ll try to get pull/scroll to refresh right. Maybe if we succeed this is interesting for the News app too. And we have the `r` shortcut for refreshing as well.

PR #1580, opened 2016-07-28
introduce very slight gradient for message header to get rid of hard border

jancborchardt:
@ChristophWurst as I showed you. Very very slight and barely noticeable, but a nice detail. ;)


ChristophWurst:
:+1:

PR #1581, opened 2016-07-28
fix messages pagination

ChristophWurst:
fixes the following bugs:
- pagination works again (next messages are added and the old ones are not removed)
- when reloading the list, some browsers scroll to the top (e.g. FF), where we now discard the first scroll event to prevent double reloading

@jancborchardt @Gomez @owncloud/mail


jancborchardt:
Nice! :+1:

PR #1582, opened 2016-07-28
add slight loading gradient, and use same spinner below

jancborchardt:
As discussed @ChristophWurst – very slight gradient behind the »check messages« spinner, and using the small spinner below as well.


ChristophWurst:
:+1: looks nicer

PR #1584, opened 2016-07-28
show only one spinner when loading the next page

ChristophWurst:
@jancborchardt as requested


ChristophWurst:
note that pagination is still kind of broken until https://github.com/owncloud/mail/pull/1581 was merged

jancborchardt:
Nice :+1:

PR #1587, opened 2016-07-29
Current message object ref

ChristophWurst:
This simply saves the message object reference instead of only the ID. This will be helpful for the features are am planning to add soon.


jancborchardt:
Seems to not entirely break anything. ;) :+1:

PR #1590, opened 2016-07-29
add shortcuts for 'flagged' and 'unseen'

ChristophWurst:
two more shortcuts of #160

@Gomez @jancborchardt @tahaalibra check it out


jancborchardt:
Works nicely! :+1:

PR #1591, opened 2016-07-29
allow users to attach local files

ChristophWurst:
client-side TODO
- [x] upload files asynchronously
- [x] add uploaded files to list of attachments
- [ ] show loading and/or progress indicator
- [ ] lock 'Send'/'Reply' button until all local attachments have been uploaded successfully
- [ ] handle errors (remove faulty attachment from list)
- [ ] add special property to attachment so that server can distinguish between local and cloud attachments 

server-side TODO
- [x] add attachment store for asynchronously uploaded files
- [ ] move uploaded attachments to temp dir (use core's temp manager)
- [ ] limit uploaded file size to a sane size (e.g. a few MBs)
- [ ] distinguish between local and cloud attachments when sending a message and load local attachments from the attachment store


mar1u5:
Would be also nice to be able to save uploaded files in your files :sweat_smile:

ChristophWurst:
> Would be also nice to be able to save uploaded files in your files

What exactly do you mean? Save the uploaded attachment in Files too? This sounds like a weird use case :-)

mar1u5:
Maybe weird ... But yes ... That is what i've meant :sweat_smile:

jancborchardt:
@mar1u5 if anything, this is something for a separate issue/PR. ;)

PR #1593, opened 2016-07-30
scroll to refresh even when already at top of list, fix #1588

### LINKED ISSUE #1588: Pull-to-refresh should ideally also work when already on top of list
Apparently the problem is that when on top of the list and scrolling, the scroll up event doesn’t fire.

jancborchardt:
Please review my beginner Backbone code :D @ChristophWurst @Gomez 

At least from testing it seems to work fine, no glitches seen yet.

### DIFF
--- a/js/views/messagesview.js
+++ b/js/views/messagesview.js
@@ -39,7 +39,7 @@ define(function(require) {
 		reloaded: false,
 		filterCriteria: null,
 		events: {
-			'wheel': 'wheel',
+			'wheel': 'onScroll',
 		},
 		initialize: function() {
 			var _this = this;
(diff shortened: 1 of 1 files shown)


jancborchardt:
Yo @owncloud/mail folks, review this awesome JS hacking I brought into the world. ;D

Gomez:
I don't get it. How to refresh? Before and after this PR?

jancborchardt:
- Before: When at the top, scroll down in the message list a tiny bit and then up again til you reach the top again.
- After: When at the top, just try scroll up more.

It's like pull to refresh on mobile.

ChristophWurst:
:-1: this triggers :100:s of requests to check messages. We need to debounce the event somehow, let me fix this

ChristophWurst:
should be good now, please test it @jancborchardt 

The only issue left is that it sometimes also reloads if one scrolls down. I assume the scroll event also fires with top === 0. Maybe we have to store the last scroll position to distinguish the direction. Could be fixed in another PR.

jancborchardt:
Nice! :+1:

PR #1600, opened 2016-08-02
update active and past contributors in the readme

ChristophWurst:
based on https://github.com/owncloud/mail/graphs/contributors
and https://github.com/owncloud/mail/graphs/contributors?from=2014-08-03&to=2016-08-02&type=c for the active ones

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,12 +35,9 @@ If you experience any issues or have enhancement suggestions you can report them
 
 ## Maintainers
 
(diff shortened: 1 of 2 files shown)


ChristophWurst:
@DeepDiver1975 @Gomez @jancborchardt @owncloud/mail objections?

jancborchardt:
Yeah, agree on this. Otherwise it makes it appear as if we do not need more help, which we do.

I would additionally compress it from a list to one comma-separated line as before. Maybe in the read me just link "and many more" to the contributor graph instead of "past contributors". . And the space before the … is missing ;)

ChristophWurst:
updated. Please take another look, thanks.

Issue #1586, opened 2016-07-29
Sometimes server error on sending from unified inbox


jancborchardt:
@ChristophWurst as we talked about, the account ID seems to not be sent so it doesn’t know which account it comes from. Only happens in unified inbox.

jancborchardt:
These are the sent parameters:
![capture du 2016-07-29 15-15-13](https://cloud.githubusercontent.com/assets/925062/17249384/0fb19b16-55a0-11e6-9abe-6df37450e7da.png)


Stacktrace:
```
Type: OCP\AppFramework\Db\DoesNotExistException
	Code: 0
	Message: Did expect one result but found none when executing: query "SELECT * FROM `*PREFIX*mail_accounts` WHERE user_id = ? and id = ?"; parameters Array
(
    [0] => jan
    [1] => 
)
; limit ""; offset ""
/home/jan/nextcloud/lib/public/AppFramework/Db/Mapper.php
Line: 289


Trace

#0 /home/jan/nextcloud/lib/public/AppFramework/Db/Mapper.php(373): OCP\AppFramework\Db\Mapper->findOneQuery('SELECT * FROM `...', Array, NULL, NULL)
#1 /home/jan/nextcloud/apps/mail/lib/db/mailaccountmapper.php(49): OCP\AppFramework\Db\Mapper->findEntity('SELECT * FROM `...', Array)
#2 /home/jan/nextcloud/apps/mail/lib/service/accountservice.php(90): OCA\Mail\Db\MailAccountMapper->find('jan', NULL)
#3 /home/jan/nextcloud/apps/mail/lib/service/unifiedaccount.php(233): OCA\Mail\Service\AccountService->find('jan', NULL)
#4 /home/jan/nextcloud/apps/mail/lib/controller/accountscontroller.php(268): OCA\Mail\Service\UnifiedAccount->resolve('15326')
#5 [internal function]: OCA\Mail\Controller\AccountsController->send(-1, 'YWxsLWluYm94ZXM...', NULL, 'Ah ok, thanks a...', '"Person name" ...', '', '', 0, '15326', NULL)
#6 /home/jan/nextcloud/lib/private/AppFramework/Http/Dispatcher.php(160): call_user_func_array(Array, Array)
#7 /home/jan/nextcloud/lib/private/AppFramework/Http/Dispatcher.php(90): OC\AppFramework\Http\Dispatcher->executeController(Object(OCA\Mail\Controller\AccountsController), 'send')
[...]

Issue #1541, opened 2016-06-24
Reply indicator only shown after refresh


jancborchardt:
Currently the reply indicator is only shown after refresh (or folder change or »check new messages«). In any case, the indicator should directly be shown as soon as the reply is sent.

ChristophWurst:
works for individual accounts, but not for the unified inbox. (I assume the existing logic needs another hack because in the unified inbox the message id is actually not the message id)

jancborchardt:
Fixed via https://github.com/owncloud/mail/pull/1604

PR #1604, opened 2016-08-03
fix unified inbox replies

ChristophWurst:
fixes https://github.com/owncloud/mail/issues/1586

@jancborchardt @owncloud/mail please test this thoroughly, I'm not too confident with all the unified inbox hacks spread across the code base…

### DIFF
--- a/js/views/message.js
+++ b/js/views/message.js
@@ -146,7 +146,7 @@ define(function(require) {
 				type: 'reply',
 				account: this.account,
 				folder: this.folder,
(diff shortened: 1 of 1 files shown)


ChristophWurst:
It now sends the unified-inbox-message-id to the server, instead of the real message id as the server otherwise fails to restore the unified-inbox information.

jancborchardt:
This works and funnily enough also solves the reply indicator being shown directly #1541! :D :+1:

PR #1607, opened 2016-08-03
fix save-draft promise/callback

ChristophWurst:
This fixes js errors when creating, but clearing a new message.

Steps to reproduce:
1. click 'New message'
2. type random content to the message content
3. wait 5sec
4. delete the content again
5. wait 5sec
6. see the js error


jancborchardt:
Works and removes the error :+1:

PR #1609, opened 2016-08-04
Search rework

ChristophWurst:
I re(thought|worked|factored) the client-side search code and UX a bit and introduced the following enhancements:
* show loading view on content instead of the *old* message list of the previously selected folder as the messages will probably disappear while searching anyway
* searching shows immediate feedback (still debouncing the actual server request though)
* search is done on an route of its own, hence the page can be refreshed
* show emtpy-config view and no-search-result view automatically (let Marionette handle it)


ChristophWurst:
@owncloud/mail @mar1u5 @MorrisJobke

jancborchardt:
As said, please use the .emptycontent class and styles. :) Then it will also make the text greyish like everywhere else.

If alignment is not perfect, please fix it in apps.css instead of here. ;)

ChristophWurst:
done

jancborchardt:
All nice except for one issue: When the »Searching for …« view is shown, there’s an unnecessary scrollbar.

ChristophWurst:
@jancborchardt should be fixed now, if not – fix it ;-)

Issue #1475, opened 2016-05-05
mail content cant be loaded


dh-connect:
### Steps to reproduce
1. start mail
2. you only see the email folder
3. right side next to the folders keeps on loading
4. no change when you click on a mail folder

### Expected behaviour
Tell us what should happen
mail should fully load

### Actual behaviour
Tell us what happens instead
only folders will be loaded

### ownCloud Mail

**ownCloud Mail version:** (see ownCloud apps admin page)
0.4.3

**Mailserver or service:** (e.g. Outlook, Yahoo, Gmail, Exchange,...)
Gmail, GMX, Web.de, own IMAP Server etc.

**Transport security - IMAP:** (None, SSL, TLS, STARTTLS) 

**Transport security - SMTP:** (None, SSL, TLS, STARTTLS)

**Number of accounts:**
10

**ownCloud Mail version build date:** (only if you are using a Nightly Build)

### ownCloud Core
9.0.2

#### Server configuration
**Operating system**:
Linux raspberrypi 4.4.8-v7+ Debian

**Web server:**
apache + php

**Database:**
mysql

**PHP version:**
PHP 7.0.5-3 (cli) 

**ownCloud version:** (see ownCloud admin page)
9.0.2

**Updated from an older ownCloud or fresh install:**
updatet from 0.4.2

**List of activated apps:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your ownCloud installation folder
```

**The content of config/config.php:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your ownCloud installation folder

{
    "system": {
        "instanceid": "xxx",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "xxx",
        ],
        "trusted_proxies": [
            "xxx",
        ],
        "datadirectory": "\/data\/ownclouddata\/ocd",
        "overwrite.cli.url": "https:\/\/xxx\/owncloud",
        "dbtype": "mysql",
        "version": "9.0.2.2",
        "dbname": "oc9",
        "dbhost": "localhost",
        "dbtableprefix": "oc_",
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "logtimezone": "UTC",
        "installed": true,
        "mail_smtpmode": "smtp",
[...]

PR #1613, opened 2016-08-05
Load accounts with index page

ChristophWurst:
* load list of accounts with index page (hidden input, recommended at http://backbonejs.org/#Collection-fetch). This reduces the time needed to load the initial data on page load.
* hide list of accounts/folders until they have finished loading and an error occurred. This also allows users to finally delete/edit accounts that are (temporarily) failing.

### DIFF
--- a/js/controller/accountcontroller.js
+++ b/js/controller/accountcontroller.js
@@ -39,6 +39,7 @@ define(function(require) {
 				Radio.navigation.trigger('setup');
 			} else {
 				var loadingAccounts = accounts.map(function(account) {
+					require('state').folderView.collection.add(account);
 					return FolderController.loadFolder(account);
 				});
 				$.when.apply($, loadingAccounts).done(function() {
--- a/js/controller/accountcontroller.js
+++ b/js/controller/accountcontroller.js
@@ -44,8 +44,7 @@ define(function(require) {
 				Radio.ui.trigger('sidebar:accounts');
 			} else {
 				var loadingAccounts = accounts.map(function(account) {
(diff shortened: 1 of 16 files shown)


js/controller/accountcontroller.js
```diff
@@ -33,23 +33,37 @@ define(function(require) {
...
+
 		$.when(fetchingAccounts).done(function(accounts) {
 			if (accounts.length === 0) {
 				defer.resolve(accounts);
 				Radio.navigation.trigger('setup');
+
+				Radio.ui.trigger('sidebar:accounts');
 			} else {
 				var loadingAccounts = accounts.map(function(account) {
+					require('state').accounts.add(account);
```
ChristophWurst: - [ ] TODO: the account might already be in the collection, hence we can remove that method call

ChristophWurst:
@jancborchardt @owncloud/mail please review!

irgendwie:
(First account is the dead account)

If one mail account fails, an unspecific message appears on top: 'Error while loading the selected account.' and both accounts appear to be dead:

deadaccounts

But after waiting another few seconds (more than 5 seconds for me) the functional account "unfolds". But the now appearing functional account is not working when trying to change folders or doing anything:

failingaccount

After removing the failing account, everything works as expected.

ChristophWurst:
> (First account is the dead account)

Known problem https://github.com/owncloud/mail/issues/1475 and unrelated to this PR :wink:

irgendwie:
Ok, then the PR works as expected :+1:

PR #1616, opened 2016-08-09
Add Signatures  Support

tahaalibra:
this will add signature support

todo
- [x] use clob in database
- [x] update composer to use signature
- [x] use service updating signature in frontend
- [x] remove files and comments not needed
- [ ] work on UI 
- [ ] remove travis errors

~~Note: This is in development, and as of current commit many things will be changed~~


tahaalibra:
what works:
- [x] backend complete
- [x] loading and saving the signatures through the account setting works

i would recommend not review this PR on current commit,
@ChristophWurst @Gomez

tahaalibra:
i have a question, when replying do we include signature?

ChristophWurst:
yes, signatures should be used for replies too

appinfo/database.xml
```diff
@@ -106,6 +106,11 @@
 				<notnull>false</notnull>
 				<length>255</length>
 			</field>
+			<field>
+				<name>signature</name>
+				<type>clob</type>
+				<notnull>true</notnull>
```
ChristophWurst: why `notnull`?

js/service/signatureservice.js
```diff
@@ -0,0 +1,90 @@
...
+
+		return defer.promise();
+	}
+
+	/**
+	 * @param accountId
+	 * @param signature
+	 * @returns {undefined}
+	 */
+	function saveAccountSignature(accountId, signature) {
```
ChristophWurst: pass an `Account` object instead, that's what we do in other services too

js/views/composer.js
```diff
@@ -527,6 +530,16 @@ define(function(require) {
 		},
 		onShow: function() {
 			this.defaultMailSelect();
+			this.showSignature();
+		},
+		showSignature: function() {
+			var alias = this.findAliasById(this.$('.mail-account').
+			find(':selected').val());
```
ChristophWurst: indent this line

js/views/composer.js
```diff
@@ -527,6 +530,16 @@ define(function(require) {
...
+			if (!this.isReply()) {
+				$('textarea[name="body"]').val('\n\n' + alias.signature);
```
ChristophWurst: use ``this.$('textarea[name="body"]')``

tests/db/mailaccounttest.php
```diff
@@ -52,7 +53,8 @@ public function testToAPI() {
 			'smtpHost' => 'smtp.marvel.com',
 			'smtpPort' => 458,
 			'smtpUser' => 'spiderman',
-			'smtpSslMode' => 'ssl'
+			'smtpSslMode' => 'ssl',
+			'signature' => 'Thanks You'
```
ChristophWurst: by default it should be ``null``

ChristophWurst:
> tahaalibra added enhancement 3 - to review in progress labels 6 days ago

Please only set the `3 - to review` if you're confident the code is ready to merge (all code written, tests fixed, no unchecked TODOs)

js/service/signatureservice.js
```diff
@@ -58,15 +58,15 @@ define(function(require) {
...
 	 * @param signature
 	 * @returns {undefined}
 	 */
-	function saveAccountSignature(accountId, signature) {
+	function saveAccountSignature(account, signature) {
 		var defer = $.Deferred();
 
 		var url = OC.generateUrl('/apps/mail/accounts/{id}/signature', {
-			id: accountId
+			id: account.accountId
```
ChristophWurst: did you test this? I assume ``account.accountId`` will be undefined as it should be a Backbone model and you have to use its `get` method instead to access the id.

mmattel:
1.) 
Are signatures differentiated between initial eMail and reply? (like you can do in Outlook)
Means:
Initial: eg long text version with much content
Reply: eg short text version with minimum content

Case needed: Business eMail need long and short signatures depending on the eMail type

2.)
When you have multiple accounts, can you have for each account a different signature set?

ChristophWurst:
> Are signatures differentiated between initial eMail and reply?

No, they aren't. I see that it would make sense, but let's keep the first version simple. Please open an enhancement ticket as soon as this feature was merged, so someone can work on that later. Thanks :-)

> When you have multiple accounts, can you have for each account a different signature set?

Yes, you can. Also for different aliases of an account.

lib/service/accountservice.php
```diff
@@ -120,12 +120,8 @@ private function buildUnifiedAccount($userId) {
 	 * @return Account[]
 	 */
 	public function updateSignature($accountId, $currentUserId, $signature) {
-		try {
-			$account = $this->mapper->find($currentUserId, $accountId);
-			$account->setSignature($signature);
-			return $this->mapper->update($account);
-		} catch(Exception $e) {
```
ChristophWurst: Hm. The service should catch implementation specific exceptions (e.g. DB exception) and throw an own exception instead. Maybe something like ``ServiceException``? Example: https://github.com/nextcloud/news/blob/master/lib/Service/ServiceException.php

tests/service/accoutservicetest.php
```diff
@@ -144,4 +144,25 @@ public function testSave() {
 		$this->assertEquals($expected, $actual);
 	}
 
+	public function testUpdateSignature() {
+		$accountId = 123;
+		$signature = "Thank You";
+
+		$this->mapper->expects($this->once())
```
ChristophWurst: is there a unit test for the mapper to ensure this method call works, too?

Gomez:
@tahaalibra Should we move this over? Someone still on it?

mariomurrent-softwaresolutions:
@tahaalibra Is your work finished? I would need this functionality asap. Maybe you can contact me.

Gomez:
@mariomurrent-softwaresolutions This is not merged. Any help is appreciated, this PR can be used as a base. 
Regarding signatures support please continue here:
https://github.com/nextcloud/mail/issues/190 

The Mail app repository was moved: 
https://github.com/nextcloud/mail

PR #1623, opened 2016-08-22
added unit tests for aliases

tahaalibra:
add unit tests for aliases

### DIFF
--- a/tests/controller/aliasescontrollertest.php
+++ b/tests/controller/aliasescontrollertest.php
@@ -31,9 +31,10 @@ class AliasesControllerTest extends TestCase {
 	private $appName = 'mail';
 	private $request;
 	private $aliasService;
-	private $userId = 'john';
+	private $userId = 'user12345';
 	private $userSession;
 	private $user;
+	private $alias;
 
 	public function setUp() {
 		$this->request = $this->getMockBuilder('OCP\IRequest')
@@ -45,7 +46,9 @@ class AliasesControllerTest extends TestCase {
 			->getMock();
 		$this->user = $this->getMockBuilder('OCP\IUser')
 			->getMock();
-
+		$this->alias = $this->getMockBuilder('\OCA\Mail\Db\Alias')
+			->disableOriginalConstructor()
+			->getMock();
 		$this->userSession->expects($this->once())
 			->method('getUser')
 			->will($this->returnValue($this->user));
@@ -54,7 +57,6 @@ class AliasesControllerTest extends TestCase {
 	}
 
 	public function testIndex() {
-		$aliases = [];
 		$accountId = 123;
 
 		$this->user->expects($this->once())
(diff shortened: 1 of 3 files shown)


tests/controller/aliasescontrollertest.php
```diff
@@ -0,0 +1,121 @@
...
+	private $user;
+
+	public function setUp() {
+		$this->request = $this->getMockBuilder('OCP\IRequest')
+			->getMock();
+		$this->aliasService = $this->getMockBuilder('OCA\Mail\Service\AliasesService')
+			->disableOriginalConstructor()
+			->getMock();
+		$this->userSession = $this->getMockBuilder('OCP\IUserSession')
+			->disableOriginalConstructor()
```
ChristophWurst: interfaces do not have a constructor :wink:

tests/controller/aliasescontrollertest.php
```diff
@@ -0,0 +1,121 @@
...
+			->getMock();
+		$this->user = $this->getMockBuilder('OCP\IUser')
+			->disableOriginalConstructor()
```
ChristophWurst: same here

tests/controller/aliasescontrollertest.php
```diff
@@ -0,0 +1,121 @@
...
+			->getMock();
+
+		$this->userSession->expects($this->once())
```
ChristophWurst: why was this assertion put in to the setUp method? Can we move that down to the individual test methods?

tests/controller/aliasescontrollertest.php
```diff
@@ -0,0 +1,121 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+use OC\AppFramework\Http;
+use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\AppFramework\Http\JSONResponse;
+use OCA\Mail\Controller\AliasesController;
+use OCP\IUserSession;
+
+class AliasesControllerTest extends PHPUnit_Framework_TestCase {
```
ChristophWurst: Please extend the core test case instead. See https://github.com/owncloud/mail/blob/afb70cf07c10b5fa22cf287b7480449da9b894b3/tests/applicationtest.php#L24 for an example

tests/db/aliasmappertest.php
```diff
@@ -0,0 +1,67 @@
...
+
+	/**
+	 * @var Alias
+	 */
+	private $alias;
+
+	/**
+	 * Initialize Mapper
+	 */
+	public function setup(){
```
ChristophWurst: would be great to have a `tearDown` method too, which restores the previous db state

tests/db/aliasmappertest.php
```diff
@@ -0,0 +1,67 @@
...
+		$db = \OC::$server->getDatabaseConnection();
+		$this->db = new Db($db);
+		$this->mapper = new AliasMapper($this->db);
+
+		$this->alias = new Alias();
```
ChristophWurst: I'd move this block down, as it's part of the test. Only initialization code should be in the setUp method in my opinion

tahaalibra:
@ChristophWurst can you review please

PR #1625, opened 2016-08-24
clean up travis file and fix PostgreSQL

ChristophWurst:
Apparently the postgresql extensions were silently removed from Travis'es php 5.4 builds, hence I changed it to use php7. This should fix failing postgre builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,19 +2,23 @@ sudo: required
 dist: trusty
 language: php
 php:
-- 5.4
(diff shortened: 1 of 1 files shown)


ChristophWurst:
@owncloud/mail @jancborchardt @DeepDiver1975 @irgendwie please review

jancborchardt:
Looks good as far as I can judge and Travis runs through :+1:

Issue #1163, opened 2015-11-21
Hardcoded /apps/ dir


Gomez:
We hardcoded the apps dir all over the place:
https://github.com/owncloud/mail/blob/master/js/require_config.js#L17

If you install mail in an alternate path with 'apps_paths' in config.php it results in 404 for js files. 

@DeepDiver1975 @PVince81 Is this correct? Is there a js way to get the apps path?

DeepDiver1975:
There is a JS way ..... I dont remember .,..

PVince81:
Have a look at `OC.appwebroots['mail']`

ChristophWurst:
@Gomez Is anybody having problems with that? Otherwise I vote for postponing to 0.4

jancborchardt:
Yeah, I’d agree we should move this to 0.3 since there hasn’t been an actual report about this.

powerpaul17:
I have the same problem with line 17 and I had to remove one "../" in order to get it working. My owncloud installation is in an "owncloud" subfolder.

juliushaertl:
fixed with #1627

PR #1627, opened 2016-08-25
Fix requirejs for other locations than apps/

juliushaertl:
Fix for https://github.com/owncloud/mail/issues/1163

This will use a dynamic baseUrl for require.js to make the mail app work within other app folders than apps/.

### DIFF
--- a/js/require_config.js
+++ b/js/require_config.js
@@ -39,6 +39,10 @@
 			}
 		}
 	});
+	// avoid optimization errors
+	requirejs.config({
+		baseUrl: OC.linkTo('mail', 'js')
+	});
 
 	require([
 		'app',


ChristophWurst:
@juliushaertl @Gomez what are the steps to test this?

juliushaertl:
@ChristophWurst You need to add a secound path to the apps_path config.php (https://github.com/nextcloud/server/blob/master/config/config.sample.php#L689-L703)

Install the mail app into the additional apps path, enable it and try to access it.

Without the patch, require.js tries to load all javascript files from "apps/mail/js/..:" but they are stored in e.g. "apps-extra/mail/js/..."

ChristophWurst:
:+1: works very nicely, thanks a lot :rocket:

PR #1628, opened 2016-08-25
inject prefilled data

ChristophWurst:
Makes the code testable and fixed the following code checker error:
```
Analysing nextcloud/apps/mail/templates/index.php
 4 errors
    line   46: OCP\User::getDisplayName - Method of deprecated class must not be called
    line   46: OCP\User::getUser - Method of deprecated class must not be called
    line   48: OCP\Config - Static method of deprecated class must not be called
    line   48: OCP\User::getUser - Method of deprecated class must not be called
```


jancborchardt:
Still works and doesn’t burn everything to the ground :+1:

PR #1629, opened 2016-08-25
replace Util::imagePath with OCP\IURLGenerator::imagePath

ChristophWurst:
Fixes
```
Analysing nextcloud/apps/mail/lib/service/htmlpurify/transformimagesrc.php
 2 errors
    line   66: Util::imagePath - Method of deprecated class must not be called
    line   75: Util::imagePath - Method of deprecated class must not be called
```
and
```
Analysing nextcloud/apps/mail/lib/service/htmlpurify/transformcssbackground.php
 1 errors
    line   68: Util::imagePath - Method of deprecated class must not be called
```


jancborchardt:
Doesn’t explode :+1:

Issue #1622, opened 2016-08-20
After upgrade keeps spinning


LEDfan:
### Steps to reproduce
1. Have a Nextcloud 9 instance
2. Upgrade to Nextcloud 10 RC 1
3. Open the mail app

### Expected behaviour
The accounts and mail should load.

### Actual behaviour
Spinners keeps spinning, accounts doesn't load.

### Mail app

**Mail app version:** 0.5.2

**Mailserver or service:** Small provider

**Transport security - IMAP:** SSL

**Transport security - SMTP:** SSL

**Number of accounts:** 3

### Server configuration
**Operating system**: Centos 7

**Web server:** Apache 2.4.5

**Database:** Mariadb 5.5.50

**PHP version:** 7.0.9

**Version:** Nextcloud 10

**Updated from an older version or fresh install:** Nextcloud 9

**List of activated apps:**

```
  - activity: 2.3.2
  - calendar: 1.3.2
  - comments: 1.0.0
  - contacts: 1.3.1.0
  - dav: 1.0.0
  - federatedfilesharing: 1.0.1
  - federation: 1.0.1
  - files: 1.5.2
  - files_external: 1.0.2
  - files_pdfviewer: 0.8.1
  - files_sharing: 1.0.0
  - files_texteditor: 2.1
  - files_trashbin: 1.0.0
  - files_versions: 1.3.0
  - files_videoplayer: 0.9.8
  - gallery: 15.0.0
  - mail: 0.5.2
  - news: 9.0.1
  - notes: 2.0.2
  - notifications: 0.3.0
  - password_policy: 1.0.0
  - provisioning_api: 1.0.0
  - richdocuments: 1.1.4
  - serverinfo: 1.1.1
  - survey_client: 0.1.5
  - systemtags: 1.0.1
  - tasks: 0.9.3
  - theming: 1.0.1
  - updatenotification: 1.0.1
  - user_external: 0.4
  - workflowengine: 1.0.0
```

**The content of config/config.php:**

<details>
```
{
    "system": {
        "instanceid": "",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "
        ],
        "datadirectory": "/srv/xxx//data",
        "dbtype": "mysql",
        "version": "9.1.0.15",
        "dbname": "nextcloud",
        "dbhost": "localhost",
        "dbtableprefix": "oc_",
        "dbuser": "***REMOVED SENSITIVE VALUE***",
[...]

PR #1634, opened 2016-08-27
delay app start until the page has been fully loaded

ChristophWurst:
fixes https://github.com/owncloud/mail/issues/1622

I don't know why this errror occurs now for so many people, but apparently the script is executed before the full page html has been loaded, therefore the element is not found.

@owncloud/mail @LEDfan @zeugmatis @irgendwie @Draghtnod review/test please.

@Gomez I think we should backport this fix and release it as 0.5.3

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -133,8 +133,10 @@ define(function(require) {
 		})(), 1000);
 	});
 
-	_.delay(function() {
+	$(function() {
+		// Start app when the page is ready
 		Mail.start();
 	});
+
 	return Mail;
 });


irgendwie:
LGTM :+1: Solves the issue in my test installation!

LEDfan:
Fixes the issue for me. Thanks!
:+1:

Gomez:
Fixed! THX!

Draghtnod:
I'm a bit late with my reply, but it's fixed for me as well. Thanks a lot!

PR #1635, opened 2016-08-27
move from tipsy to tooltip

jancborchardt:
Finally get rid of all those errors :)

Please review @owncloud/mail


irgendwie:
* Added a commit to fix - tooltip has no "gravity", instead "placement" is used
* tooltip has no live option, therefore removed it

Now the tooltips should be at the correct position.

